PR #20, opened 2019-02-24
Symbol arrays

samphippen:


src/rubyfmt.rb
```diff
@@ -1824,26 +1813,53 @@ def format_program(line_metadata, sexp, result)
...
+  end
 end
 
 def main
   file_data = File.read(FILE)
   file_data = file_data.gsub("\r\n", "\n")
 
   line_metadata = extract_line_metadata(file_data)
-  sexp = Ripper.sexp(file_data)
+  sexp = Parser.new(file_data).parse
```
kddeisz: Heads up because I just ran into this with prettier, make sure to check `parser.error?` so that you can bail out if ripper failed to generate the AST

PR #22, opened 2019-03-05
use ARGF so we read from stdin

tenderlove:
I will add a test for this, just busy rn.


tenderlove:
(I want to run rubyfmt like `cat out.txt | ruby src/rubyfmt.rb`)

PR #40, opened 2019-03-13
Allow stdin to be read as well as file

sentientmonkey:
This allows for easier integration with plugins.

I also have a vim-plugin working with this change 😃 

https://github.com/sentientmonkey/vim-rubyfmt


sentientmonkey:
Oops... didn't need this after all. nevermind!

PR #69, opened 2019-03-23
Fix trailing comma in super call.

### LINKED ISSUE #68: Breaks on trailing comma in super call
* Ruby version: ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin18]
* Rubyfmt git sha: 046d0b96447f9adb244a82586bcb5b8c1f5a852c

## Input file

Minimal example:

```ruby
super(1,)
```

Real world example: https://github.com/thoughtbot/gitsh/blob/master/lib/gitsh/tab_completion/dsl/parser.rb#L79-L82

## Rubyfmt's output

```
Traceback (most recent call last):
        7: from ./rubyfmt.rb:2448:in `<main>'
        6: from ./rubyfmt.rb:2445:in `main'
        5: from ./rubyfmt.rb:2290:in `format_program'
        4: from ./rubyfmt.rb:2290:in `each'
        3: from ./rubyfmt.rb:2291:in `block in format_program'
        2: from ./rubyfmt.rb:2284:in `format_expression'
        1: from ./rubyfmt.rb:2235:in `block in format_expression'
./rubyfmt.rb:1445:in `format_super': nope args (RuntimeError)
```

## This breaks because

A trailing comma on the list of arguments to `super`.

georgebrock:
This commit transforms the subtree for a super call's arguments with a trailing comma to look like the subtree we would have got for arguments without a trailing comma.

This might be an overly-naive approach to fixing this problem, but if nothing else the fixture is hopefully useful.

Fixes #68


samphippen:
@georgebrock so it turns out that trailing commas in arg like things have the same behaviour everywhere, so I just refactored out my fix for normal method calls for this one, but thanks for the initial pass!

georgebrock:
@samphippen I thought that might be the case, but wasn't familiar enough to find your existing fix and avoid duplicating it.

Glad the patches were helpful. I'll run some of the other thoughtbot open source repos through rubyfmt and see if I can find more edge cases.

PR #74, opened 2019-03-24
Add script to test rubyfmt against any Git repo.

georgebrock:
This commit adds a script to run all of the Ruby files in some external Git repo through rubyfmt, only outputting errors and the paths to the files where they were encountered.

The purpose of the script is to make it easier to throw a whole bunch of code at rubyfmt, so that we can more easily find bugs and uncovered edge-cases.

The issues and PRs I've opened so far were found by running a bunch of thoughtbot repos through rubyfmt (so far: gitsh, factory_bot, and an internal Rails app called Hub), and I've found enough stuff that it seemed useful to turn the awful shell one-liner I was using into a properly reusable script.

### DIFF
--- a/scripts/test_repo.sh
+++ b/scripts/test_repo.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+if [[ "$#" -ne 1 ]]
+then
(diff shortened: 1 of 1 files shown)


samphippen:
Sure why not.

PR #85, opened 2019-03-25
Support range literals with no upper bound.

georgebrock:
This syntax is only valid in 2.6, so the tests unsurprisingly fail on 2.3 and 2.5.

Not sure if it's better to throw away these fixtures, or change the CI setup to run against 2.6.


samphippen:
I need to rejiggle the CI script, which is currently hideously naieve.

fixtures/range_with_no_upper_bound_actual.rb
```diff
@@ -0,0 +1 @@
+(1..)
```
teoljungberg: I guess we could place this under `fixtures/2.6` to only have it run under 2.6?

georgebrock: Hi, Teo! I think that makes sense, but it's also the reason we'll need to change the test script: `fixtures/2.6` is currently gitignored, because the current test script copies `fixtures/2.5` to `fixtures/2.6`.

samphippen:
push incoming.

samphippen:
oh joy there's a failing CI on 2.3 :(

samphippen:
@georgebrock ignoring the failing CI, WDYT of the bash patch?

samphippen:
docker.

samphippen:
@georgebrock this is now blocked by #91

georgebrock:
@samphippen The script changes look good to me.

Nice use of bc(1) for the version comparison, I was thinking of doing something awful with `Gem::Version.new(RUBY_VERSION)`.

samphippen:
@georgebrock please provide final LGTM, I got this green on all 3 rubys we CI.

samphippen:
I just did a mega rebase and I want to see if I didn't break anything.

scripts/test.sh
```diff
@@ -1,14 +1,29 @@
...
+echo $RUBY_VERSION
+if [[ `echo "2.5<=$RUBY_VERSION" | bc -l` -ne 0 ]]
+then
+    test_folder fixtures/2.5
+fi
+
+if [[ `echo "2.6<=$RUBY_VERSION" | bc -l` -ne 0 ]]
+then
+    test_folder fixtures/2.6
+fi
```
georgebrock: What do you think about making this automatically discover directories for new versions? Would probably make it easier for contributors to add version-specific tests.

This works for me on macOS:

```shell
for dir in $(find fixtures -type d -name '2.*')
do
    fixture_version=$(basename "$dir")
    if [[ $(echo "$fixture_version<=$RUBY_VERSION" | bc -l) -ne 0 ]]
    then
        test_folder $dir
    fi
done
```

samphippen: Please open a PR for this, right now I'd like to go on a merge spree.

PR #86, opened 2019-03-26
Use `git diff --no-index` over `diff(1)`

teoljungberg:
It gives contributors a more familiar interface with the diffs the test
produce.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -6,4 +6,9 @@ for file in `ls fixtures/*_expected.rb` `ls fixtures/$(ruby -v | grep -o '\d\.\d
 do
     time ruby --disable=gems src/rubyfmt.rb `echo $file | sed s/expected/actual/` > /tmp/out.rb
     diff -u /tmp/out.rb $file
+    if [[ $? -ne 0 ]]
+    then
+        echo "got diff"
+        exit 1
+    fi
 done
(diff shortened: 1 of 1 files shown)


scripts/test.sh
```diff
@@ -5,10 +5,5 @@ cp -r fixtures/2.5 fixtures/2.6
 for file in `ls fixtures/*_expected.rb` `ls fixtures/$(ruby -v | grep -o '\d\.\d')/*_expected.rb`
 do
     time ruby --disable=gems src/rubyfmt.rb `echo $file | sed s/expected/actual/` > /tmp/out.rb
-    diff /tmp/out.rb $file
-    if [[ $? -ne 0 ]]
-    then
-        echo "got diff"
-        exit 1
-    fi
```
georgebrock: This conditional will exit the script early when a test fails. Is this an intentional change? Seems distinct from changing the output format.

scripts/test.sh
```diff
@@ -5,10 +5,5 @@ cp -r fixtures/2.5 fixtures/2.6
...
+    git diff --no-index /tmp/out.rb $file
```
georgebrock: I think `diff -u` (or `diff --unified`) will give the same output format as `git diff --no-index`, and seems more clear to me because we wouldn't be invoking Git to diff two untracked files.

scripts/test.sh
teoljungberg: In my head I thought

> `diff` will exit with 1 regardless, so this is moot.

But you're right - we want to keep the short circuit. Re-added in 3384d50.

scripts/test.sh
teoljungberg: I thought since we were in the confounds of a git repository, `git diff --no-index` made sense. But I can see your point about clarity and can agree.

Changed to `diff -u` in 6f2d4d8

PR #87, opened 2019-03-26
String refactor

samphippen:


src/rubyfmt.rb
```diff
@@ -2457,11 +2458,84 @@ def on_heredoc_end(*args, &blk)
...
+  ONE_SLASH_IN_TSTRING = "\\"
+
+  # Rubyfmt renders all string literals as double quotes for consistency's sake
+  # this method fixes up individual tstring content items to ensure that they
+  # represent the same string literal once reformatted. This method is not
+  # called on string literals that were already surrounded by double quotes,
+  # and so this formatting only applies to tstring_contents from strings that
+  # need to be manipulated to correctly reform as a double quoted string
+  #
+  def fixup_tstring_content_for_double_quotes(string, is_single_quote:, delimiter:)
```
georgebrock: I don't see any obvious changes we could make to reduce the complexity of this method, but I wondered when working on this bug if eventually splitting it into an explicit “unescape all the things” step followed by a “escape the relevant things” step could give us two smaller complex things instead of one big one?

src/rubyfmt.rb
```diff
@@ -2477,17 +2551,27 @@ def on_string_literal(*args, &blk)
...
+        is_single_quote =  next_string.start_with?("%q")
         (args[0][1..-1] || []).each do |part|
           next if part.nil?
           case part[0]
           when :@tstring_content
-            fixup_tstring_content_for_double_quotes(part[1])
+            fixup_tstring_content_for_double_quotes(
+              part[1],
+              is_single_quote: is_single_quote,
+              delimiter: quote,
```
georgebrock: `quote` might be the wrong thing to pass here, because in some cases the opening and closing delimiters are different (e.g. `%q(` and `)`) and in those cases we need to unescape closing delimiters, not opening ones.

We can push the closing quotes to a stack in `#on_tstring_end`. We could even use the same stack, and pop twice to get the distinct quotes:

```ruby
next_string_end = @string_stack.pop
next_string_start = @string_stack.pop
```

src/rubyfmt.rb
```diff
@@ -2477,17 +2551,27 @@ def on_string_literal(*args, &blk)
...
           case part[1][0]
           when :@tstring_content
-            fixup_tstring_content_for_double_quotes(part[1][1])
+            fixup_tstring_content_for_double_quotes(
+              part[1][1],
+              is_single_quote: true,
+              delimiter: quote,
+            )
           else
             raise "got non tstring content in single string"
```
georgebrock: Do we need to guard against this? Without a serious Ripper bug, I'm not sure how we could get a `string_embexpr` in the wrong type of string. If we don't need this guard clause, we can probably collapse the `/'/` and `/\W/` branches into one.

src/rubyfmt.rb
samphippen: can you provide a sample string that you think might surface this bug and check it?

georgebrock: I thought a string like this would be a problem, but I was wrong:

```ruby
%<foo\>>
```

I was mixing up `quote` (which didn't change in this PR) with `next_string` (which is new, and came from the stack).

I think things would be more clear if we grabbed the opening and closing quotes in the same way, but there is no bug.

src/rubyfmt.rb
samphippen: do you think the high amount of comments and well named constants makes this more understandable?

paging @jcoglan for review thoughts on how we could de-complex this.

src/rubyfmt.rb
samphippen: This is definitely a "is the ruby parser well written" test, but I'd like to keep it in case ruby ever does weird things in the future.

src/rubyfmt.rb
```diff
@@ -2477,17 +2551,27 @@ def on_string_literal(*args, &blk)
...
           end
         end
-      elsif [")", "}", "|"].include?(quote)
+      elsif /\W/ === quote && quote != "\""
```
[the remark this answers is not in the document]
samphippen: 👍

src/rubyfmt.rb
samphippen: I could go either way on this, how strongly do you feel I should maintain both in the stack?

georgebrock: I feel more strongly about the names: `next_string` and `quote` doesn't express to me how closely related they are.

I feel less strongly about how we find the two things, but I find an `#on_tstring_end` method easier to read than `[@file_lines[lineno-1].bytes[column-1]].pack("c*")`

src/rubyfmt.rb
jcoglan: So the naming does make it clear what's being done here, but the things that make me nervous about this code is that you're working on bare strings, and the order of transformations appears to matter, including the fact you set up placeholders in early steps that are then removed later.

I've no idea to what extent this is possible, and in ruby it might be more pain than it's worth, but I was wondering whether you could tokenise the string first, and work on transforming that token stream, rather than using `gsub` which has the effect of introducing characters that have special meaning to later operations.

This would remove the need for placeholders, since each transformation would leave anything it didn't recognise untouched. It might also facilitate doing some of the "nested" transformations where you set some placeholders, do another transformation, then return to what you were originally doing. This is the sort of things blocks are great at.

The only caveat I have is that ruby doesn't have pattern-matching and so the code for this method may end up being a nightmare to read.

src/rubyfmt.rb
```diff
@@ -611,7 +611,14 @@ def format_params(ps, params, open_delim, close_delim)
   rest_params = params[3] || []
   more_required_params = params[4] || []
   kwargs = params[5] || []
+
   kwrest_params = params[6] || []
+  # on ruby 2.3 this position contains literally the integer 183 if a `**` is
+  # given in the splatted kwargs position. Why, I have no idea.
+  if kwrest_params == 183
```
jcoglan: I am scraeming

PR #89, opened 2019-03-26
Extract expressions hash to reduce run time by 5-20%

toreriklinnerud:
Will you take an easy speedup?

```bash
bench "ruby --disable=gems src/rubyfmt.rb ~/.gems/activerecord-5.2.2.1/lib/active_record/reflection.rb"
```

**Before**
```
time                 258.7 ms   (255.6 ms .. 260.4 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 256.8 ms   (253.4 ms .. 258.4 ms)
std dev              3.020 ms   (580.4 μs .. 4.400 ms)
variance introduced by outliers: 16% (moderately inflated)
```

**After**
```
time                 210.6 ms   (203.2 ms .. 216.9 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 215.5 ms   (212.3 ms .. 219.3 ms)
std dev              4.790 ms   (3.543 ms .. 6.407 ms)
variance introduced by outliers: 14% (moderately inflated)
```

TL;DR: On a hairy Rails file, 258.7 ms => 210.6 ms


src/rubyfmt.rb
```diff
@@ -2272,6 +2272,96 @@ def format_keyword_with_args(ps, expression, keyword)
...
+  :@cvar => lambda { |ps, rest| format_cvar(ps, rest) },
+  :mlhs_paren => lambda { |ps, rest| format_mlhs_paren(ps, rest) },
+  :mrhs_add_star => lambda { |ps, rest| format_mrhs_add_star(ps, rest) },
+  :while => lambda { |ps, rest| format_while(ps, rest) },
+  :lambda => lambda { |ps, rest| format_lambda(ps, rest) },
+  :rescue_mod => lambda { |ps, rest| format_rescue_mod(ps, rest) },
+  :xstring_literal => lambda { |ps, rest| format_xstring_literal(ps, rest) },
+  :@backref => lambda { |ps, rest| format_backref(ps, rest) },
+  :@CHAR => lambda { |ps, rest| format_character_literal(ps, rest) },
+}
```
samphippen: can you try adding `freeze` here to see if it goes faster still?

toreriklinnerud: Added, not a measureable improvement but doesn't hurt.

Issue #98, opened 2019-03-29
Fails against TRICK source code


mame:
* Ruby version: 2.6.0
* Rubyfmt git sha: `e12b794`

## Input file

* https://github.com/tric/trick2018/blob/master/01-kinaba/entry.rb
* https://github.com/tric/trick2018/blob/master/02-mame/entry.rb
* https://github.com/tric/trick2018/blob/master/05-tompng/entry.rb
* https://github.com/tric/trick2018/blob/master/07-colin/entry.rb
* https://github.com/tric/trick2018/blob/master/08-tompng/entry.rb

## Rubyfmt's output

For example, run against 01-kinaba/entry.rb:

```
$ ruby /tmp/rubyfmt.rb entry.rb 
Traceback (most recent call last):
	18: from /tmp/rubyfmt.rb:2557:in `<main>'
	17: from /tmp/rubyfmt.rb:2554:in `main'
	16: from /tmp/rubyfmt.rb:2397:in `format_program'
	15: from /tmp/rubyfmt.rb:2397:in `each'
	14: from /tmp/rubyfmt.rb:2398:in `block in format_program'
	13: from /tmp/rubyfmt.rb:2391:in `format_expression'
	12: from /tmp/rubyfmt.rb:2323:in `block in <main>'
	11: from /tmp/rubyfmt.rb:1370:in `format_unless_mod'
	10: from /tmp/rubyfmt.rb:1378:in `format_conditional_mod'
	 9: from /tmp/rubyfmt.rb:2391:in `format_expression'
	 8: from /tmp/rubyfmt.rb:2351:in `block in <main>'
	 7: from /tmp/rubyfmt.rb:1947:in `format_alias'
	 6: from /tmp/rubyfmt.rb:171:in `with_start_of_line'
	 5: from /tmp/rubyfmt.rb:1948:in `block in format_alias'
	 4: from /tmp/rubyfmt.rb:2391:in `format_expression'
	 3: from /tmp/rubyfmt.rb:2315:in `block in <main>'
	 2: from /tmp/rubyfmt.rb:1089:in `format_symbol_literal'
	 1: from /tmp/rubyfmt.rb:171:in `with_start_of_line'
/tmp/rubyfmt.rb:1098:in `block in format_symbol_literal': didn't get ident in right position [[:@kw, "BEGIN", [1, 9]]] (RuntimeError)
```

## Note

Thank you for the great tool.

[...]

PR #99, opened 2019-03-29
Mame fixes

samphippen:


samphippen:
This seems to fix all but one of the issues in #98

PR #101, opened 2019-03-29
Support for loops.

georgebrock:
Real world example: https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/fog.rb#L104


samphippen:
do we have to :P ?

LGTM

src/rubyfmt.rb
```diff
@@ -2238,6 +2246,38 @@ def format_while(ps, rest)
...
+
+  ps.emit_indent if ps.start_of_line.last
+
+  ps.emit_for
+  ps.emit_ident(" ")
+  format_list_like_thing_items(ps, [loop_vars], true)
+  ps.emit_ident(" ")
+  ps.emit_in
+  ps.emit_ident(" ")
+  ps.with_start_of_line(false) do
```
georgebrock: Not sure this is necessary, but it seemed appropriate to follow the same structure as while loops.

samphippen: if you've ever emitted tokens and not a newline, it is good practice to throw one of these blocks in.

Issue #88, opened 2019-03-26
Lint shell files using shellcheck(1)


teoljungberg:
I noticed that the shell files that is leveraged by the test tooling done by `make` aren't shellcheck compatible. I'd gladly do the work to convert the `.sh` files into a compatible state and add a proposed `make lint` command that would run on CI.

Would that be anything of value this early in the projects state?

samphippen:
I'd support adding this as a linting pass if it goes in our dockerfiles without too much difficulty

samphippen:
@teoljungberg do you want to have a go at updating the various CI dockerfiles to include this?

teoljungberg:
@samphippen - sure. Would you prefer having it baked in or having a dedicated "linting" image that's run as a prior step?

samphippen:
@teoljungberg let's throw it in all 3 of the containers for now.

samphippen:
This is merged! Yay

PR #108, opened 2019-03-30
Lint `.sh` files using `shellcheck(1)`

teoljungberg:
References #88

Changes:
- Add `shellcheck(1)` linting to Docker images.
- Fix linter warnings from `shellcheck(1)`. The following warnings were
  fixed:

  https://shellcheck.com/wiki/SC1009
  https://shellcheck.com/wiki/SC1020 
  https://shellcheck.com/wiki/SC1072
  https://shellcheck.com/wiki/SC1073 
  https://shellcheck.com/wiki/SC2001
  https://shellcheck.com/wiki/SC2006 
  https://shellcheck.com/wiki/SC2044
  https://shellcheck.com/wiki/SC2045 
  https://shellcheck.com/wiki/SC2046
  https://shellcheck.com/wiki/SC2053 
  https://shellcheck.com/wiki/SC2086
  https://shellcheck.com/wiki/SC2094 
  https://shellcheck.com/wiki/SC2162
  https://shellcheck.com/wiki/SC2181 
  https://shellcheck.com/wiki/SC2230

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -5,4 +5,5 @@ if [ $RES != "puts(\"a\")"]
 then
     exit 1
 fi
+./scripts/lint.sh
 ./scripts/test.sh
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -ex
-RES=`echo "puts 'a'" | ruby src/rubyfmt.rb`
-if [ $RES != "puts(\"a\")"]
+
+RES=$(echo "puts 'a'" | ruby src/rubyfmt.rb)
+
+if [ "$RES" != "puts(\"a\")" ]
 then
     exit 1
 fi
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-git grep -l "bin\/bash" | xargs shellcheck
+git grep -l "bin\/bash" | xargs shellcheck -f gcc
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+git grep -l "bin\/bash" | xargs shellcheck
--- a/scripts/test_repo.sh
+++ b/scripts/test_repo.sh
@@ -12,6 +12,7 @@ for file in $(git --work-tree="$repo_path" --git-dir="$repo_path/.git" ls-files 
 do
   full_path="$repo_path/$file"
   errors=$(ruby --disable=gems src/rubyfmt.rb "$full_path" 2>&1)
+  # shellcheck disable=SC2181
   if [[ "$?" -ne 0 ]]
   then
     echo "$full_path"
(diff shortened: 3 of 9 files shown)


ci/run.sh
```diff
@@ -1,8 +1,11 @@
...
-RES=`echo "puts 'a'" | ruby src/rubyfmt.rb`
-if [ $RES != "puts(\"a\")"]
+
+RES=$(echo "puts 'a'" | ruby src/rubyfmt.rb)
+
+if [ "$RES" != "puts(\"a\")" ]
 then
     exit 1
 fi
+./scripts/lint.sh
```
teoljungberg: I added the linting step before the test script is run since it's much faster.

scripts/lint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+git grep -l "bin\/bash" | xargs shellcheck -f gcc
```
teoljungberg: Do we want to lint all files ending in `.sh` instead.

teoljungberg: The `gcc` formatter makes `:make` inside of `vim(1)` pickup the linting errors.

If we want to use the default formatter, I'll happily change this to use it.

scripts/test_repo.sh
```diff
@@ -12,6 +12,7 @@ for file in $(git --work-tree="$repo_path" --git-dir="$repo_path/.git" ls-files
 do
   full_path="$repo_path/$file"
   errors=$(ruby --disable=gems src/rubyfmt.rb "$full_path" 2>&1)
+  # shellcheck disable=SC2181
```
teoljungberg: I disabled this lint since I didn't come up with a way to run a command, and save its output.

@georgebrock might have a trick up his sleeve.

teoljungberg:
I have one odd test failure that only occurs on this branch, and not on master.

```
+ diff /tmp/out.rb fixtures/nested_conditionals_expected.rb
+ read -r file
++ echo fixtures/2.6/range_with_no_upper_bound_expected.rb
++ sed s/expected/actual/
+ ruby --disable=gems src/rubyfmt.rb fixtures/2.6/range_with_no_upper_bound_actual.rb
src/rubyfmt.rb:2744:in `main': undefined method `error' for #<Parser:0x000055b77e3bce38> (NoMethodError)
        from src/rubyfmt.rb:2753:in `<main>'

real    0m0.047s
user    0m0.010s
sys     0m0.020s
make: *** [docker-test] Error 1
```

Ideas?

georgebrock:
> I have one odd test failure that only occurs on this branch, and not on master.

That failure looks to me like Ruby can't parse the input file, which probably means that a 2.6-specific example is running with an older-than-2.6 Ruby. Not sure why _that_ would be happening though.

scripts/test_repo.sh
georgebrock: I don't know another way of doing this, aside from the one I already used in the original script.

samphippen:
this branch has started running 2.6 fixtures on ruby 2.5 for some reason. @teoljungberg can you rebase master and take a look at what's going on? It looks like this is executing all the tests in fixtures recursively, not just the all ruby version fixtures.

teoljungberg:
@samphippen Resolved in dbdbf08.

The bug happened since `ls PATTERN` didn't do so recursively as `find(1)` did, adding a `-maxdepth 1` ensured only the correct ruby fixtures were used.

samphippen:
merged via 6ecfc65011072549142a525cd896c9f671213b1c

PR #110, opened 2019-03-30
Use eval for string escaping.

georgebrock:
Even after Sam's serious improvements to string handling in #87, it still kinda bothered me that we had to do a bunch of work that Ruby must already know how to do.

I spent too much time crawling around in `parse.y` figuring out how Ruby handles character escaping with different string delimiters and, long story short, there doesn't seem to be a way of getting at that information from Ripper without making changes to `parse.y`. Maybe that's worth doing sometime, but since we want to support old Rubies too it wouldn't help immediately.

Still, the Ruby interpreter knows how to properly unescape the content of string literals. So here's a wildly different approach: Instead of trying to parse and re-escape all possible strings ourselves, or coax the information out of Ripper, let's delegate the responsibility to Ruby by using `#eval` and `#inspect` to get a correctly escaped double-quoted string for any input string. Effectively this is doing what @jcoglan suggested and tokenising the string first, it's just cheating by using Ruby's implementation of that code instead of writing it ourselves.

I think this code is easier to understand, and less likely to be hiding surprising bugs. If nothing else, it's a lot shorter.

Couple of things that still bother me:

1. **Performance** might take a hit, which is important. I ran `time ./scripts/test.sh` and `time ./scripts/test_repo.sh ../gitsh` a few times against this branch and against master and didn't see a clear winner, but more thorough benchmarking might be needed.
1. **Security** is always a worry with `#eval`, although there's no immediately obvious attack that I can see here. Given the input data is escaped strings from Ripper, there would have to be a bug in Ruby string parsing to exploit this, and even then an attacker would already need sufficient access to a machine to get that malicious string into a developer's code so that their editor would `rubyfmt` over it.

### DIFF
--- a/benchmarks/Gemfile
+++ b/benchmarks/Gemfile
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
+# gem "rails"
+gem "benchmark-ips"
--- a/benchmarks/Gemfile.lock
+++ b/benchmarks/Gemfile.lock
@@ -1,0 +1,13 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    benchmark-ips (2.7.2)
+
+PLATFORMS
+  ruby
+
+DEPENDENCIES
+  benchmark-ips
+
+BUNDLED WITH
+   2.0.1
--- a/benchmarks/string_escaping/bench.rb
+++ b/benchmarks/string_escaping/bench.rb
@@ -1,0 +1,33 @@
+require 'benchmark/ips'
+require 'ripper'
+
+$: << File.dirname(__FILE__)
+
+require 'parser_eval'
+require 'parser_munging'
+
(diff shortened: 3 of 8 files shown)


samphippen:
iiiiiiiiiiiiiiiiiiiiiiiiiinteresting.

samphippen:
@georgebrock I'll hit this with benchmark ips and see what the difference is. Thanks!

samphippen:
the good news is that these bench the same:

```
Warming up --------------------------------------
             munging     6.000  i/100ms
                eval     6.000  i/100ms
           munging 2    44.000  i/100ms
              eval 2    46.000  i/100ms
Calculating -------------------------------------
             munging     65.748  (± 4.6%) i/s -      1.644k in  25.064874s
                eval     67.743  (± 3.0%) i/s -      1.698k in  25.085068s
           munging 2    461.021  (± 5.0%) i/s -     11.528k in  25.088951s
              eval 2    456.908  (± 5.9%) i/s -     11.408k in  25.066459s

```

I think we should do this because the implementation is radically simpler.

Issue #107, opened 2019-03-29
rubyfmt is not idempotent for all inputs


georgebrock:
* Ruby version: ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin18]
* Rubyfmt git sha: b918aaeda3d7c9213fd1d6c6509d698085d66f2c

## Input file

A line extracted from `fixtures/command_call_raise_actual.rb`:

```
raise ArgumentError.new bees
```

## Rubyfmt's output

`raise ArgumentError.new bees` becomes `raise ArgumentError.new(bees)` on the first pass through rubyfmt, and then `raise(ArgumentError.new(bees))` on the second pass.

After the second pass the output is stable.

```
$ cat test.rb
raise ArgumentError.new bees
$ cat test.rb | ruby src/rubyfmt.rb
raise ArgumentError.new(bees)
$ cat test.rb | ruby src/rubyfmt.rb | ruby src/rubyfmt.rb
raise(ArgumentError.new(bees))
$ cat test.rb | ruby src/rubyfmt.rb | ruby src/rubyfmt.rb | ruby src/rubyfmt.rb
raise(ArgumentError.new(bees))
```

## This breaks because

If someone were running rubyfmt automatically on every save, it ought to make no changes to parts of the file that hadn't been modified since the last save.

georgebrock:
I'm going to close this: after #117 and #112, rubyfmt is idempotent for all examples in the fixtures directory, and as that test suite grows we'll continue to check the new examples.

PR #112, opened 2019-03-31
Test idempotence

georgebrock:
This PR updates the test script to run the expected fixtures through `rubyfmt` and compare the output to the original file. The intention is to catch idempotence bugs, i.e. situations where running a file through `rubyfmt` a second time would change the file.

The failures seem to be around whether or not we include parens on a method call. I started looking at fixing them, but Ruby apparently has many, _many_ ways of representing the same call, and it's more than I'll have time to fix today.

Current failures are:

- ~~`fixtures/command_call_raise_expected.rb`~~ (fixed by #117)
- ~~`fixtures/nested_command_expected.rb`~~ (fixed by #117)
- ~~`fixtures/until_while_mod_expected.rb`~~ (fixed on this branch)

Related to #107

### DIFF
--- a/scripts/test_fixtures.sh
+++ b/scripts/test_fixtures.sh
@@ -9,7 +9,15 @@ test_folder() {
 
         if ! diff -u /tmp/out.rb "$file"
         then
-            echo "got diff"
+            echo "got diff between formated actual and expected"
+            exit 1
+        fi
+
+        time ruby --disable=gems src/rubyfmt.rb "$file" > /tmp/out.rb
+
+        if ! diff -u /tmp/out.rb "$file"
+        then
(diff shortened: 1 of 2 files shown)


samphippen:
@georgebrock are you going to take a crack at this or should I?

georgebrock:
@samphippen I intend to take a crack at it later in the week, but it's too big for the time I have today. If you want to tackle it sooner, feel free. I'll check here for comments before I do anything.

samphippen:
all yours, just checking you weren't expecting me to deal with it!

georgebrock:
Two of these will be fixed by #117, which leaves one more that is unrelated to method calls.

georgebrock:
@samphippen I think this is ready to go.

PR #117, opened 2019-04-02
Normalisation experiment

georgebrock:
What?
=====

Ripper has a _lot_ of ways it can represent method calls. They include:

- `method_add_arg`, for any call with parens
- `command`, for a bare method call without parens
- `command_call`, for dotted method call without parens
- `call`, for a dotted method call without parens or arguments
- `vcall`, for a bare method call without parens or arguments

There are also some very method-call-like things, where the same formatting
rules probably ought to apply. They are:

- `zsuper`, for super with no arguments
- `super`, for super with arguments, or with empty parens
- `yield`
- `return`

This commit normalises these s-expressions into the following structure
before formatting them:

    [
      :method_call,
      receiving_object,
      method_name,
      original_used_parens,
      arguments,
    ]

Where:

- `receiving_object` is an array of expressions, which can be concatenated
  to produce the receiving object. This could include operators like `.`,
  `&.`, or `::`.

- `method_name` is an expression representing the name of the method.

- `original_used_parens` is a boolean, indicating if the original source
  code used parens. This is useful for `super` calls, where empty parens
  change the semantics.

- `arguments` is an array or expressions representing the arguments.

To simplify further, a few other s-expressions are also modified. In the
`receiving_object` array:

- Operators are represented as `[:dot, expression_or_symbol]`. This fits
  neatly with the existing `#format_dot` method.

- Call-like things are represented as `[:keyword, keyword]`, e.g. `[:keyword, "return"]`.

And in the `arguments` array:

- Splats are represented as `[:spat, expression]` where `expression` is
  the splatted argument.

- Blocks passed as params are represented as `[:to_proc, expression]`
  where `expression` is the block argument.

Now that these cases are all handled consistently, some fixtures have
necessarily changed to make sure they're consistent with each other.

Why?
====

When a subtle change to the formatting of a method call (e.g. the addition
or parentheses) could change the s-expression representation of the
expression (e.g. from a `command` to a `method_add_arg`), and therefore
change the code path followed by rubyfmt, it was easy for differences
between formatting to creep in

There are a couple of reasons this matters:

1. We want rubyfmt to be idempotent, i.e. running a file through it many
   times should produce the same output as running it through once.
   Maintaining that property while formatting the same kind of thing in
   multiple places would be difficult.

2. We're likely to go back-and-forth on what the output should look like.
   This means a single place to make changes to method calls and other
   call-like structures will be beneficial.

### DIFF
--- a/fixtures/cannibalization_4_expected.rb
+++ b/fixtures/cannibalization_4_expected.rb
@@ -1,1 +1,1 @@
-Metadata.relative_path File.expand_path(rerun_path)
+Metadata.relative_path(File.expand_path(rerun_path))
--- a/fixtures/cannibalization_4_expected.rb
+++ b/fixtures/cannibalization_4_expected.rb
@@ -1,1 +1,1 @@
-Metadata.relative_path(File.expand_path(rerun_path))
+Metadata.relative_path File.expand_path(rerun_path)
--- a/fixtures/command_paren_expected.rb
+++ b/fixtures/command_paren_expected.rb
@@ -1,1 +1,1 @@
-@proc_keys.merge(proc_keys_from(metadata))
+@proc_keys.merge proc_keys_from(metadata)
--- a/fixtures/command_paren_expected.rb
+++ b/fixtures/command_paren_expected.rb
@@ -1,1 +1,1 @@
-@proc_keys.merge proc_keys_from(metadata)
+@proc_keys.merge(proc_keys_from(metadata))
--- a/fixtures/fib_expected.rb
+++ b/fixtures/fib_expected.rb
@@ -13,7 +13,7 @@ end
 
 def fibs
   10.times do |i|
-    p(fib(i))
+    p fib(i)
   end
 end
 
--- a/fixtures/fib_expected.rb
+++ b/fixtures/fib_expected.rb
@@ -13,7 +13,7 @@ end
 
 def fibs
   10.times do |i|
-    p fib(i)
+    p(fib(i))
   end
 end
 
--- a/fixtures/heredoc_ending_with_curly_expected.rb
+++ b/fixtures/heredoc_ending_with_curly_expected.rb
@@ -1,4 +1,4 @@
-raise(<<-EOM.gsub(/^\s+\|/, ""))
+raise <<-EOM.gsub(/^\s+\|/, "")
               |#{"*" * 50}
               |:#{key} is not allowed
               |
--- a/src/rubyfmt.rb
+++ b/src/rubyfmt.rb
@@ -2246,6 +2246,36 @@ def format_keyword(ps, rest)
   ps.emit_ident(rest[0])
 end
 
+def use_parens_for_method_call(method, args, original_used_parens, method_call_depth)
+  # Always use parens for the shorthand `foo::()` syntax
+  return true if method == :call
+
+  # Never use parens for some methods and keywords
+  return false if ["require", "return", "raise"].include?(method[1])
+
+  # Follow the original code style for super and yield
+  # Note that `super()` has different semantics to `super`
+  return original_used_parens if ["super", "yield"].include?(method[1])
+
+  # No parens if there are no arguments
+  return false if args.empty?
+
+  # Use parens if this call appears while formatting another call
+  # e.g. in a chain like `previous.this_one(1).next`, or in an argument
+  # like `another_call(this_one(1))`
+  return true if method_call_depth > 0
+
+  # Don't use parens if this call only has one argument, and that
+  # argument is another call
+  return false if (
+    args.length == 1 &&
+    (args[0][0] == :method_call && args[0][4].length > 0 || args[0][0] == :method_add_block)
+  )
+
+  # If in doubt, use parens
+  true
+end
+
 def format_method_call(ps, rest)
   ps.emit_indent if ps.start_of_line.last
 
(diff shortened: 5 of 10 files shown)


src/rubyfmt.rb
```diff
@@ -1998,10 +1810,6 @@ def format_alias(ps, expression)
   ps.emit_newline if ps.start_of_line.last
 end
 
-def format_field(ps, rest)
```
georgebrock: Not sure I should have removed this, but it didn't cause any tests to fail.

src/rubyfmt.rb
```diff
@@ -2401,8 +2209,251 @@ def format_redo(ps, expression)
...
+end
+
+def format_method_call(ps, rest)
+  ps.emit_indent if ps.start_of_line.last
+
+  chain, method, original_used_parens, args = rest
+
+  args = args.map { |a| normalize(a) }
+
+  use_parens = (
```
georgebrock: This is pretty awful, but it's as close as I could get to the logic that existed across the various `format_*` methods that are replaced by this method.

src/rubyfmt.rb
```diff
@@ -2401,8 +2209,251 @@ def format_redo(ps, expression)
...
+def format_keyword(ps, rest)
+  ps.emit_ident(rest[0])
+end
+
+def format_method_call(ps, rest)
+  ps.emit_indent if ps.start_of_line.last
+
+  chain, method, original_used_parens, args = rest
+
+  args = args.map { |a| normalize(a) }
```
georgebrock: This is kind of a hack: the args would get normalised anyway when we pass them to `format_expression`, but by pre-normalising them we can check if any of them are a `:method_call` instead of having to check for all the different Ripper variants on method calls.

georgebrock:
Hmm, this breaks the RSpec stress test, so definitely not ready to merge!

fixtures/cannibalization_4_expected.rb
```diff
@@ -1 +1 @@
-Metadata.relative_path(File.expand_path(rerun_path))
+Metadata.relative_path File.expand_path(rerun_path)
```
samphippen: why did these parens go missing?

fixtures/command_paren_expected.rb
```diff
@@ -1 +1 @@
-@proc_keys.merge(proc_keys_from(metadata))
+@proc_keys.merge proc_keys_from(metadata)
```
samphippen: same question

fixtures/fib_expected.rb
```diff
@@ -13,7 +13,7 @@ def fib(i)
 
 def fibs
   10.times do |i|
-    p(fib(i))
+    p fib(i)
```
samphippen: same question

src/rubyfmt.rb
samphippen: I have something here.

samphippen:
how does this work in the case that a call receiver is like

`(Foo::Bar::(nil.object_id)).bees(1)`

fixtures/cannibalization_4_expected.rb
georgebrock: They became consistent with [`nested_command_expected.rb`](https://github.com/samphippen/rubyfmt/blob/master/fixtures/nested_command_expected.rb) and [`command_call_raise_expected.rb`](https://github.com/samphippen/rubyfmt/blob/master/fixtures/command_call_raise_expected.rb). Same with the other fixtures that changed.

We could go the other way: change `nested_command_expected.rb` to use more parens and tweak `use_parens` to never use parens with `raise`.

src/rubyfmt.rb
georgebrock: This does cause the RSpec stress test to fail. Bringing it back.

georgebrock:
> how does this work in the case that a call receiver is like
>
> `(Foo::Bar::(nil.object_id)).bees(1)`

I'd completely forgotten about the `foo::()` shorthand. Fixed in abd5d3f.

fixtures/cannibalization_4_expected.rb
georgebrock: Alternative in 3cfae79 which is more consistent with the previous behaviour. Still changes a couple of fixtures, but all of these parens are back.

src/rubyfmt.rb
georgebrock: I pulled it out into a method, but further improvements would be great.

fixtures/heredoc_ending_with_curly_expected.rb
```diff
@@ -1,4 +1,4 @@
-raise(<<-EOM.gsub(/^\s+\|/, ""))
+raise <<-EOM.gsub(/^\s+\|/, "")
```
georgebrock: This changes because the current logic is “never use parens with `raise`”. For `super` and `yield` we currently follow whatever the original source did, and we could also do that for `raise` if we didn't want to change this fixture.

samphippen:
@georgebrock  please rebase against master, which now has RSpec ci

bquorning:
I was wondering if the normalisation logic doesn’t belong in the `Parser` class, e.g. by implementing an `#on_command` method instead of the `#normalize_command` method? And would that even be possible to implement?

georgebrock:
@bquorning Good question! I suspect the code might be less clear if we did it using parser events, for two reasons:

- The current separation makes normalisation a clearly separate thing from either parsing or formatting. If it was mixed in with parsing or with formatting, it could be harder to figure out what's what. Currently the custom code we have in the parser is all concerned with getting at information Ripper doesn't provide by default (e.g. unescaped string values).

- I _think_ an event-driven approach would mean maintaining more state, which could make it harder to debug the normalisation code because we'd have to consider not only the current event but also the state that previous events have put us in.

I haven't worked with Ripper a whole lot (my commits in this project are it), so I could be making the wrong assumptions.

georgebrock:
Green tests!

The failure was coming from this expression: `require(path_for(formatter_ref)) ? retry : raise`

I'd made “no parens on `require`” too broad. I think there's a thing to figure out about what the rules for using parens should really be in practice, but at least this no longer changes the semantics of RSpec Core.

bquorning:
Ah, got it. I had misread the code and didn’t see it was a separate step.

Would having e.g. a `Formatter` and a `Normalizer` class (or module) make the code clearer? It’s a bit confusing having every method defined without a namespace. (Probably only a Normalizer class would be within the scope of this PR)..

samphippen:
@georgebrock I need to revisit where we should have parens and shouldn't. Could you build an example ala `string_literals_stress_test.rb` with as many variants as you can think of with md5 puts output based verification?

samphippen:
@georgebrock my intent is to merge this once CI is green.

Issue #121, opened 2019-04-04
Autoformat big integers


samphippen:
Rubyfmt should automatically do convenience formatting on big integers:

```ruby
5
500
5000
50000000
```

should become

```ruby
5
500
5_000
50_000_000
```

samphippen:
The code you want to update is here: https://github.com/samphippen/rubyfmt/blob/master/src/rubyfmt.rb#L719

twe4ked:
Rubocop is probably a good starting place for this.

https://github.com/rubocop-hq/rubocop/blob/f7559c18dc8911c82585bc8bae7e4de03edc5a5e/lib/rubocop/cop/style/numeric_literals.rb

mikker:
Of course, there are always exceptions eg. credit card numbers which would make more sense as `4111_1111_1111_1111`. Not necessarily something needing to be _fixed_ but wanted to mention.

robotdana:
another exception is for example rails' schema.rb version which is YYYY_MM_DD_HHMMSS

I'd prefer it if rubyfmt passed through numbers with underscores unchanged if it can

twe4ked:
#123 was closed with the following comment:

> I think that given that people like to put their underscores in all sorts of interesting places, we shouldn't do this. […]

PR #123, opened 2019-04-05
Autoformat big integers with underscore separator

azach:
Been following rubyfmt for a bit, thought I'd take a stab at #121. Feels like a straight single iteration over the string array is the quickest and most predictable approach (as opposed to regex et. al). Note that unlike Rubocop, this operates on integer digits of length 4, as well as negative integers.


src/rubyfmt.rb
```diff
@@ -716,6 +716,25 @@ def format_int(ps, rest)
   ps.emit_indent if ps.start_of_line.last
 
   int = rest[0]
+
+  if int[0] != "0"
+    int_array = []
+    counter = 0
+
+    (0..int.length - 1).each do |i|
+      next if int[int.length - 1 - i] == "_"
```
bquorning: I’m not sure how important performance is yet, but I fiddled a bit with your code to reduce array lookups:
```rb
  if int[0] != "0"
    int_array = []
    counter = 0

    int.chars.reverse.each do |char|
      next if char == "_"

      if counter > 0 && counter % 3 == 0
        int_array.unshift("_")
      end

      int_array.unshift(char)
      counter += 1
    end

    int = int_array.join
  end
```
Simple benchmarking suggests the fixtures file is handled around 45% faster, and exclusively working on small integers works 15% faster.

azach: Nice suggestion -- I benchmarked against some options (including `push` vs `unshift`) initially, but it definitely seems like this is a good improvement.

src/rubyfmt.rb
```diff
@@ -716,6 +716,25 @@ def format_int(ps, rest)
   ps.emit_indent if ps.start_of_line.last
 
   int = rest[0]
+
+  if int[0] != "0"
+    int_array = []
+    counter = 0
+
+    int.chars.reverse.each do |char|
+      next if char == "_"
```
teoljungberg: I'd use `#reject` to remove the chars that include `_` before looping around with `#each`.

But I'm unsure how @samphippen feels about that.

src/rubyfmt.rb
```diff
@@ -716,6 +716,25 @@ def format_int(ps, rest)
   ps.emit_indent if ps.start_of_line.last
 
   int = rest[0]
+
+  if int[0] != "0"
+    int_array = []
```
teoljungberg: Rather than keeping an array and a counter, could the each block be expressed as a `reduce` instead?

src/rubyfmt.rb
azach: Definitely could do that, I was just trying to optimize doing this in a single iteration over the string.

src/rubyfmt.rb
azach: Currently, the counter is needed to keep track of how many digits we've placed since a valid Ruby integer of length N can contain up to N-1 underscores (e.g. 123 vs 1_2_3), though if we combine it with the rejection of underscores as proposed above it's certainly possible. However, then we need to determine the placement of the underscore based on the length of the string, which might make the logic a little less clear?

samphippen:
Does `int.chars.reject { |x| x == "_" }.reverse.each_slice(3).map(&:join).to_a.reverse.join("_")` work?

kddeisz:
Not to throw a wrench in things, but there are occasionally intentionally formatted integers (https://github.com/rails/rails/blob/master/activerecord/lib/active_record/schema_dumper.rb#L57-L62). Maybe a check if it's already got underscores would be good?

samphippen:
fiosfjqweiofasdklgj;adrihejqfoiewfoqwifjqoiwfjqweifojpqqwfioqwpjfwqef.

Yeah ok, @azach let's leave the string alone if they're already there

azach:
Good wrench @kddeisz, and that does pose an interesting problem. 

If you have an integer, say `10000`), then it gets formatted to `10_000`, and you want to modify it further (to say, `100000000`) you either have to manually modify the new int correctly, or remove the existing underscores to let the formatter do its magic again -- neither one is a particularly great experience, since if you're fiddling with the code formatting to get it to format correctly it somewhat defeats the purpose. @samphippen thoughts?

samphippen:
@georgebrock WDYT?

src/rubyfmt.rb
```diff
@@ -716,6 +716,12 @@ def format_int(ps, rest)
   ps.emit_indent if ps.start_of_line.last
 
   int = rest[0]
+  int_chars = int.chars
+
+  if int_chars[0] != "0" && !int_chars.detect { |c| c == "_" }
```
teoljungberg: If we want to be fancy, we could use `none?` rather than `![].detect`.

georgebrock: We should probably go with `!int_chars.include?("_")` — it's less code than `#detect` or `#none`, and something like an order of magnitude faster.

src/rubyfmt.rb
```diff
@@ -716,6 +716,12 @@ def format_int(ps, rest)
...
+    int = int.chars.reverse.each_slice(3).map(&:join).join("_").reverse
```
georgebrock: We could reuse the `int_chars` variable here, instead of calling `int.chars` again

samphippen:
@georgebrock there's an issue where if we have autoformatted to `10_000`, and then a user adds another 0, if we leave user underscores alone, then  we no longer can format that.

bquorning:
Just a thought: This looks like a very isolated feature, which could easily be turned on and off with a configuration flag.

bestie:
Just a data point: When working on e-commerce systems I liked to format currency like `1_000_00` for one thousand Euro. Wiping that out wouldn't stop me using rubyfmt, it's a very small readability optimization I'd be prepared to trade-off.

samphippen:
I think that given that people like to put their underscores in all sorts of interesting places, we shouldn't do this. Closing. @azach thanks for at least getting this started!

PR #124, opened 2019-04-06
First pass of respecting user blanklines

samphippen:
Makes rubyfmt detect on and then respect single user blanklines, see the many churned fixtures for examples.

Hard part here was keywords and parenish things, which doesn't necessarily get locators in the ruby parse tree, but I hacked it with line properties.

@georgebrock WDYT?


samphippen:
closing in favour of a PR that combines this and something else!

PR #129, opened 2019-04-08
Fix a couple of minor inconsistencies

bquorning:
Trying to read my way through the source code, and found two seemingly unused methods plus a couple of minor inconsistencies.

PR #134, opened 2019-04-14
Re apply #129 on top of previous commits

samphippen:
Re applies #129 on top of the PR


samphippen:
LMAO this isn't going apply cleanly now, so I'm just going to close this and instead we'll focus on making these improvements to the new code.

PR #141, opened 2019-06-12
Break hashes

samphippen:


src/requires.rb
```diff
@@ -3,6 +3,8 @@
 require "ripper"
 require "stringio"
 
+MAX_WIDTH=100
```
mikker: `ENV['RUBYFMT_MAX_WIDTH']` _*runs away*_

PR #158, opened 2020-03-06
Reorganize Test Fixtures

### LINKED ISSUE #157: Reorganize fixtures + test scripts
Right now we have a monolithic `fixtures/` directory with versions under it for ruby version specific features.

It's worth noting **most** of the fixtures are tiny (<20 lines of code), but at least one (`fixtures//rspec_core_notifications_{actual, expected}.rb`) is large (coming directly from rspec).

I'd like to split fixtures in to `small` (syntax units) and `large` (entire files from projects that we've formatted and hand inspected the formatting of).

It's also worth noting some fixture files are unpaired (they have an actual, but not an expected, or the other way around).

This issue is fixed when:

* the fixtures directory is further broken in to `small`, and `large`
* there are test scripts to run small on it's own, large on it's own, and both
* both small and large have the ability to have version subdirectories run by the scripts
* there is a script to validate all fixtures are paired

coreyja:
Fixes #157 

This PR reorganizes the test fixtures, as well as the test fixtures script.

- Reorganizes the tests such that
  - There is a small and large first level directory. These sort the files into buckets that will be able to be run independently
- Added/modified the following scripts
  - test_fixtures.sh tests all fixtures
  - test_fixtures_small.sh tests small fixtures (everything but rspec)
  - test_fixtures_large.sh tests large fixtures (rspec)
  - lint_fixtures.sh makes sure all the files have a match, and there aren't extra incorrectly named ruby files floating around in the fixtures dir


script/test_fixtures.sh
```diff
@@ -1,42 +1,61 @@
...
-    find "$1" -name "*_expected.rb" -maxdepth 1 | while read -r file
-    do
-        # shellcheck disable=SC2001
-        time ruby --disable=gems rubyfmt.rb "$(echo "$file" | sed s/expected/actual/)" > /tmp/out.rb
+    current_dir="$1"
+
+    ## Check that both files or neither exist
+    if [ -f "$current_dir/expected.rb" ] || [ -f "$current_dir/actual.rb" ]; then
+      [ -f "$current_dir/expected.rb" ] || (echo "$current_dir is missing the expected.rb file" && exit 1)
+      [ -f "$current_dir/actual.rb" ] || (echo "$current_dir is missing the actual.rb file" && exit 1)
```
coreyja: This currently checks that each file has a pair, in the main script. Do you like this or do you want a separate lint script that can check this?

script/test_fixtures.sh
```diff
@@ -1,42 +1,61 @@
...
+        cat /tmp/diff.out
+        exit 1
+      fi
+
+      ## Test if the formatting is idempotent
+      time ruby --disable=gems rubyfmt.rb "$current_dir/expected.rb" > /tmp/out.rb
+      if ! diff -u /tmp/out.rb "$current_dir/expected.rb" > /tmp/diff.out
+      then
+        echo "got diff between formatted expected and expected (not idempotent)"
+        cat /tmp/diff.out
```
coreyja: I also captured the diff output and only output on a failure to keep less in the terminal. The rust still logs a lot and I didn't capture that so this could definitely be removed if wanted

penelopezone:
this looks good! Couple questions:

1. did you run shellcheck over the shell files? That'll eventually be part of CI
2. would you be able to add scripts to test just the small fixtures, and just the large fixtures
3. would you be able to break the `_actual` `_expected` matching in to it's own script?

coreyja:
1. I believe it did, but will double check it's passing
2. I did this via an argument passed to the main script. You can pass a directory to it and it will only run that subset. So `test_fixtures.sh small` or even `test_fixtures.sh small/alias`. If you want a separate script for small/large I can definitely do that too!
3. For sure will work on that when I have another spare cycle or two

Any complaints or thoughts about how I strctures the fixtures directory? With subfolders for each year instead of the filename prefix? Just making sure you're cool with that too before I more forward. I think it seems easier to maintain this way but totally up to you

penelopezone:
@coreyja I'd rather have the files organised as they are now, at the same directory level, because that way an alphabetical sort will show the pairs next to each other, not with each example in a subfolder. This is mostly a function of how I work with the fixtures directory.

As for the scripts, if you could break it into separate scripts, testing all, small, and large, that'd be great.

script/test_methods.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-source ./scripts/functions.sh
```
coreyja: Noticed while running `lint.sh` that this was plural while the dir was not so changed

script/lint.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 set -exou pipefail
-git grep -l "bin\/bash" | xargs shellcheck -x -f gcc
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+
+git grep -l "bin\/bash" | xargs shellcheck -x -f gcc --source-path "$GIT_ROOT"
```
coreyja: The `--source-path` change makes it so this script isn't dependant on the cwd for the purposes of finding the sourced files.

This DOES however require a relatively new version of shellcheck (I had to upgrade locally to get this option). So if that doesn't work or isn't required I can remove this too

coreyja: Removed this, cause I realized other scripts needs to be run repo root anyways for other reasons, so no a super strong need for this

But if you actually do want it just let me know lol

coreyja:
Ok! All files pass shellcheck via the `lint.sh` script in the repo!

We have a couple scripts now (and an added function in `functions.sh`)

- `test_fixtures.sh` tests all fixtures
- `test_fixtures_small.sh` tests small fixtures (everything but rspec)
- 'test_fixtures_large.sh` tests large fixtures (rspec)
- 'lint_fixtures.sh` makes sure all the files have a match, and there aren't extra incorrectly named ruby files floating around in the `fixtures` dir

I also removed a few random things I had previously added (like only outputting the git diff on failure) so as to not try to optimize for my workflow when I'm not the one actually using this most lol. I can add anymore things like that you do eventually want, but thought I'd keep it simpler for now

penelopezone:
@coreyja I manually merged this! Thank you so much

PR #163, opened 2020-03-31
Make it faster

sgrif:
Prior to this change, of the time spent in rubyfmt's code, about 60% of
that was in deserialization. I suspected (and turned out to be correct)
that most of that was just the cost of constructing this giant AST,
which we can't really reduce without a more fundamental redesign.
However, I did notice that we were spending a lot of time in some serde
internal code, so I looked into that.

Most of our enums use `#[serde(untagged)]`, which tries to deserialize
each variant and returns the first one that succeeds. This is partially
because serde doesn't support the special kind of internally tagged
representation that Ripper uses, and partially because a lot of our
enums are only sometimes tagged.

Serde generates the best code it can for the general case here. Since it
can't know whether the deserializer actually consumes its input when it
deserializes something, the deserializer can only be used once. But we
need to attempt deserialization multiple times. To do this, Serde
deserializes into an enum which has one variant per method on
`serde::de::Visitor`. The issue here is for `visit_seq`, the appropriate
type to use is a `Vec`, and we are *mostly* dealing with arrays, so this
does a lot of unneccessary allocation and copying (the copy is the main
issue here, jemalloc is making the allocation mostly free).

However, we *are* using a deserializer that deserialize multiple times.
We're just looking at a pointer to a Ruby object, we're not doing any
mutation, and that pointer can be copied for free.

So this commit adds a custom derive which tightly couples our
deserialize impls to our deserializer, by trying to deserialize a
`VALUE`. I've written a `Deserialize` impl which *should* fail if it
ever works with any other deserializer (the way the trait is structured
prevents me from statically enforcing that it only works with one
deserializer). Once we have the `VALUE`, we can construct our
deserializer from it, and try to deserialize as many times as we want.

The code we generate is nearly identical to the code generated by
`#[serde(untagged)]`. The main difference is that we're deserializing
into `VALUE` instead of a `serde::private::de::Content`, and using our
deserializer instead of `serde::private::de::ContentRefDeserializer`. If
for some reason this change needs to be reverted in the future, all
cases of `#[derive(RipperDeserialize)]` can be replaced with
`#[derive(Deserialize)] #[serde(untagged)]` with no change in behavior.
Mostly.

There is one notable change here, which is that any struct which
consisted only of its tag, where it was a tuple struct of a single
element tuple of the tag had to have that wrapping tuple removed. It's
not clear to me if this is a bug in serde, or if its doing some
intentional magic for single element tuple structs. But I've approached
this change from multiple angles and every time I've done anything that
affected the codegen I had to remove the wrapping tuple.

Additionally, by "fixing" this, the expression variant `DotCall`, which
was not properly handled before, was now appearing instead of `Call`
(since it techncially matched). I've swapped the order so we always try
`Call` first, and removed the newtype so it correctly represents the
`:call` tag directly. While this structure still isn't handled, I
believe it's closer to what was intented to work.

On my machine this resulted in a roughly 10% performance increase. This
is what I was expecting but is still sadly low for the amount of time
this took. Luckily when I removed the `to_json` call, I forgot to remove
the `require "json"` so I did that in this commit as well so I can
pretend that it's actually a 20% perf increase.


mikker:
Would you also consider making it _harder, better, stronger_?

PR #170, opened 2020-04-06
Make defs work with memoist-style definitions

fables-tales:
/cc @samandmoore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean clippy lint
+.PHONY: clean clippy lint fmt
 
 debug: target/rubyfmt_debug.bundle
 
@@ -29,3 +29,6 @@ target/release/librubyfmt.a: native/src/*.rs native/Cargo.toml
 
 clean:
 	rm -rf target/
+
+fmt:
+	cd native && cargo fmt && git add -u ./
--- a/fixtures/small/memoist_actual.rb
+++ b/fixtures/small/memoist_actual.rb
@@ -1,0 +1,7 @@
+memoize def foo
+end
+
+memoize def foo(x, y, z)
+  aboiqjfdw
+  foo(bar, 1, 2)
+end
--- a/fixtures/small/memoist_expected.rb
+++ b/fixtures/small/memoist_expected.rb
@@ -1,0 +1,11 @@
+memoize(
+  def foo
+  end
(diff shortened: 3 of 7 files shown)


samandmoore:
Yay! Lemme know when you want me to try to break this thing some more @penelopezone 😁

timriley:
It's nice to see that these method "annotations" will work, thank you @penelopezone!

If you don't mind, I have a question about this PR — is the structure demonstrated in `memoist_expected.rb`:

- Just a temporary state, demonstrating that things are at least _working_ right now, but with a view to having rubyfmt eventually preserve the original structure shown in `memoist_actual.rb`?
- Or is it the way you want rubyfmt to format these things for the long run?

Thank you!

penelopezone:
@timriley nothing is final right now. That said:

* This style of method definition is pretty uncommon (I'd literally never seen it before @samandmoore showed me the example, and I've been doing Ruby for > 10 years).
* Rubyfmt is a "prefers no special cases" formatter
* Rubyfmt always prints parentheses on method calls
* I think this form would be pretty ugly:
```ruby
memoize(def foo
end)

so, I'm disinclined to special case definitions of this form right now, but I am open to changing my mind.

**Comment by timriley** [comment]
Thanks for explaining your thinking, @penelopezone 🙏 

So I've done that `memoize def foo` thing a fair bit – specifically with memoizing helper gems – ever since the Ruby 2.1 change for method definitions to return a symbolised name, mostly because I liked the concision of it.

I recognise it's perhaps a somewhat unusual approach (evidenced even by how it can break the syntax highlighting in certain editors), but I'm happy to offer this as an anecdotal data point for you at least.

If a tool like rubyfmt (and I'm super keen to adopt it) wound up re-formatting them to this structure:

```ruby
memoize(
  def foo(x,y,z)
    # ...
  end
)
```

Then I think I'd have to give up the approach entirely, since the concision has gone out the window by that point!

I'm definitely not saying my one piece of feedback should force you to change the approach, but I do think it's a nice little piece of expressiveness we lose by not having this special-cased (though I also realise this is a slippery slope to go down - I'll leave it with you from here).

FWIW I think I've also seen Avdi use this "method annotation style" with `private def foo(x,y,z)` too... IIRC you and he might've had a Twitter convo about it.

PR #172, opened 2020-04-07
Onboard rspec expectations

penelopezone:


native/src/format.rs
```diff
@@ -1547,12 +1553,8 @@ pub fn format_defs(ps: &mut ParserState, defs: Defs) {
...
 
         ps.emit_dot();
-        match ident_or_kw {
-            IdentOrKw::Ident(ident) => format_ident(ps, ident),
-            IdentOrKw::Kw(kw) => {
-                handle_string_and_linecol(ps, kw.1, kw.2);
-            }
-        }
+        let i = ident_or_kw.to_def_parts();
+        handle_string_and_linecol(ps, i.0, i.1);
```
sgrif, line 1557: What do you think about adding some names here?

```suggestion
        let (ident, linecol) = ident_or_kw.to_def_parts();
        handle_string_and_linecol(ps, ident, linecol);
```

native/src/format.rs
```diff
@@ -463,6 +464,9 @@ pub fn use_parens_for_method_call(
     }
 
     if name == "return" || name == "raise" || name == "yield" || name == "break" {
+        if ps.current_formatting_context() == FormattingContext::Binary {
```
sgrif, line 467: What do you think about adding a method for this to avoid coupling to the structure?

```suggestion
        if ps.current_formatting_context().is_binary() {
```

And the method would look like

```
fn is_binary(&self) -> bool {
    match self {
        FormattingContext::Binary => true,
        _ => false,
    }
}
```

PR #174, opened 2020-04-07
Use `#[serde(flatten)]` where possible

sgrif:
https://github.com/serde-rs/serde/pull/1771 adds support for
`#[serde(flatten)]` on tuple structs. This removes any manual
`Deserialize` impls that were only needed because of the lack of support
for this.

This points at git to get the branch, we can point back at released
serde once that PR is merged.

### DIFF
--- a/native/src/ripper_tree_types.rs
+++ b/native/src/ripper_tree_types.rs
@@ -474,42 +474,11 @@ pub enum StringContentPart {
 }
 
 def_tag!(string_content_tag, "string_content");
-#[derive(Debug, Clone)]
-pub struct StringContent(pub string_content_tag, pub Vec<StringContentPart>);
-
-impl<'de> Deserialize<'de> for StringContent {
-    fn deserialize<D>(deserializer: D) -> Result<StringContent, D::Error>
-    where
-        D: Deserializer<'de>,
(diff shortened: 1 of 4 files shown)


native/src/ripper_tree_types.rs
```diff
@@ -474,42 +474,11 @@ pub enum StringContentPart {
...
-            }
-        }
-
-        deserializer.deserialize_seq(StringContentVisitor)
-    }
-}
+#[derive(Deserialize, Debug, Clone)]
+pub struct StringContent(
+    pub string_content_tag,
+    #[serde(flatten)] pub Vec<StringContentPart>,
```
penelopezone, line 480: can the flatten annotations go on the line above, as opposed to the same line?

sgrif: rustfmt says no

sgrif: I want to remove the tags so this will eventually be on a single line anyway. And/or I'll change it to a named struct in which case it'll go on its own line

sgrif:
Going to close this since I doubt serde with that change in the next day and it'll be easier to just redo this once it's released

PR #179, opened 2020-04-10
Cli

penelopezone:


rubyfmt.rb
```diff
@@ -205,8 +204,37 @@ def on_regexp_literal(*args)
...
-Rubyfmt::format_to_stdout(file_data, parsed)
-STDOUT.close
+
+def parse_file(fn)
+  file_data = File.read(fn)
+  [file_data, Parser.new(file_data).parse]
+end
+
+def rubyfmt_dir(dn)
+  files = Dir[File.join(dn, "**/*.rb")]
```
robotdana, line 214: you may like to also find Gemfile, Capfile, .gemspec, *.rake, config.ru, etc

robotdana: also files with no extension but with ruby shebang (SORRY)

penelopezone: If you'd like, I'd welcome a patch that supports this usecase, but for right now I'm going to merge this.

PR #189, opened 2020-04-17
More ruby files

robotdana:
From this comment: https://github.com/penelopezone/rubyfmt/pull/179#discussion_r407064119

More files use ruby than just .rb. This is as exhaustive as i could make
it with a little bit of research (all the files i've come across as a
ruby developer, plus whatever documented variations there are).

*.builder is from rails (https://guides.rubyonrails.org/action_view_overview.html#builder)
*.jbuilder is from rails (https://guides.rubyonrails.org/action_view_overview.html#jbuilder)
Gemfile is from bundler (https://bundler.io/man/gemfile.5.html)
Capfile, capfile is from capistrano (https://github.com/capistrano/capistrano/blob/master/lib/capistrano/application.rb#L5)
Guardfile, .Guardfile is from guard (https://github.com/guard/guard/blob/master/README.md#start)
*.ru is from rack (https://github.com/rack/rack/wiki/(tutorial)-rackup-howto)
Rakefile, rakefile, Rake, *.rake is from rake (https://github.com/ruby/rake/blob/master/doc/command_line_usage.rdoc)
.irbrc, irb.rc, _irbrc, $irbrc is from irb (https://docs.ruby-lang.org/en/2.7.0/IRB.html#module-IRB-label-Configuration)
.pryrc, pryrc is from pry (https://github.com/pry/pry/wiki/Pry-rc#overview)

Also checks ruby files with no extension, but with a ruby shebang

This is largely based on code from
https://github.com/robotdana/fast_ignore which in turn is based on code
from stdlib find but with fewer system calls so it's very slightly faster.

You may want just the first commit, as that's significantly less ruby.

### DIFF
--- a/rubyfmt.rb
+++ b/rubyfmt.rb
@@ -210,24 +210,29 @@ def parse_file(fn)
   [file_data, Parser.new(file_data).parse]
 end
 
+RUBY_PATHS_PATTERN = "**/{*.rb,*.builder,*.jbuilder,*.ru,*.rake,*.gemspec,Gemfile,Capfile,capfile,Guardfile,.Guardfile,rakefile,Rakefile,Rake,.irbrc,irb.rc,_irbrc,$irbrc,pryrc,.pryrc,.simplecov}"
 def rubyfmt_dir(dn)
-  files = Dir[File.join(dn, "**/*.rb")]
-  files.each do |glob_fn|
-    file_data, parsed = parse_file(glob_fn)
-    Rubyfmt::format_to_file(glob_fn, file_data, parsed)
+  Dir.glob(File.join(dn, RUBY_PATHS_PATTERN), File::FNM_DOTMATCH).each do |glob_fn|
+    rubyfmt_file(glob_fn)
   end
 end
 
+def rubyfmt_file(fn)
+  file_data, parsed = parse_file(fn)
+  Rubyfmt::format_to_file(fn, file_data, parsed)
+end
+
 first = ARGV.shift
+
 if first == "-i"
   ARGV.each do |fn|
     if File.directory?(fn)
       rubyfmt_dir(fn)
     else
-      file_data, parsed = parse_file(fn)
-      Rubyfmt::format_to_file(fn, file_data, parsed)
+      rubyfmt_file(fn)
     end
   end
+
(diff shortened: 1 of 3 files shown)


rubyfmt.rb
```diff
@@ -205,34 +205,74 @@ def on_regexp_literal(*args)
...
+  def self.===(error)
+    error.is_a?(ArgumentError) && error.message.match?(RE)
   end
 end
 
+RUBY_PATHS_PATTERN = "**/{*.rb,*.builder,*.jbuilder,*.ru,*.rake,*.gemspec,Gemfile,Capfile,capfile,Guardfile,.Guardfile,rakefile,Rakefile,Rake,.irbrc,irb.rc,_irbrc,$irbrc,pryrc,.pryrc}"
+SHEBANG_PATTERN = /\A#!.*\bruby\b/
+def rubyfmt_dir(path)
+  Dir.new(path).each do |child|
+    next if child == '.' || child == '..'
```
robotdana, line 227: if this project doesn't run on 2.4 or earlier then this line could be removed and the previous changed to `Dir.each_child do |child|`

robotdana:
you know what? I missed some https://github.com/rubocop-hq/rubocop/blob/c4ce01c333f962165576031e0ab8d61a4aafa8ed/config/default.yml#L11

penelopezone:
Hi @robotdana,

so I think instead of having a massive include list in rubyfmt proper, I think we should  include Gemfile, all the rails file types, and *.rb, and anything with a shebang by default. Anything additional, users should add to a file called ".rubyfmtrc" or something

robotdana:
Do you currently have a plan for this file containing other things? or should i just assume a text file list of file name patterns with no headings or etc? and look in pwd and home for .rubyfmtrc? pwd overriding or merging with home?

(i suddenly understand your no configuration kick)

so we're on the same page, rails types are *.rb *.rake *.builder *.jbuilder and *.ru?

penelopezone:
I will think about this and get back to you

fables-tales:
hi @robotdana please could you re-open this against the new primary branch, I rebased the entire git history and changed the branch and that force closed all the PRs

PR #192, opened 2020-04-24
Collapse rendering to a single target stack

fables-tales:
This finalizes collapsing rendering to a single stack of rendering
targets, harmonizing the "render queue", which is just a wrapper around
a Vec of tokens, and a BreakableEntry, which has some special state
management for dealing with line splitting of things like arrays,
hashes, method calls, and method definitions.

This moves all the behaviour of knowing which target we're touching out
of the ParserState and in to the new RenderTargetStack struct.

Things I like about this:

* ParserState is smaller, and it's kind of our god object
* We're operating on fewer primitives, and more structs with named
  methods

Things I don't like about this:
* RenderTargetStack knows a lot about whether or not it's currently
  targetting a breakable entry, but maybe that's fine
* We have to expose pushing and popping a BreakableEntry to the
  ParserState, it'd be nice if target_stack could have a
  `with_breakable_entry` function, instead of exposing direct push and
  pop methods, but I couldn't work out how to make the borrow checker
  play nice with it

There is a very minor behaviour change in this commit in that if
BreakableEntries contain hard newlines they will flush comments, but
this didn't break any of the fixtures (largely because we don't have
fixtures for that behaviour). I'm going to fix that up in a future
commit.


sgrif:
> it'd be nice if target_stack could have a
with_breakable_entry function, instead of exposing direct push and
pop methods, but I couldn't work out how to make the borrow checker
play nice with it

If you have the code you tried and/or the error you got I can probably help with this

penelopezone:
> If you have the code you tried and/or the error you got I can probably help with this

I'll push up a separate branch

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
...
+    contents: Vec<LineToken>,
+}
+
+fn insert_at<T>(idx: usize, target: &mut Vec<T>, input: &mut Vec<T>) {
+    let drain = input.drain(..);
+    let mut idx = idx;
+    for item in drain {
+        target.insert(idx, item);
+        idx += 1;
+    }
```
sgrif, line 15: Every call to insert has to potentially realloc, and then copy all elements after it. What do you think of something like this:

```suggestion
    let mut target_tail = input.split_off(idx);
    target.append(input);
    target.append(&mut target_tail);
```

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
+use crate::comment_block::CommentBlock;
+use crate::line_tokens::LineToken;
+
+#[derive(Debug, Clone)]
```
sgrif, line 4:
```suggestion
#[derive(Default, Debug, Clone)]
```

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
...
+        let mut a2 = a1.clone();
+        a1.insert(1, 4);
+        super::insert_at(1, &mut a2, &mut vec![4]);
+        assert_eq!(a1, a2);
+    }
+}
+
+impl LineTokenCollection {
+    pub fn new() -> Self {
+        LineTokenCollection { contents: vec![] }
```
sgrif, line 32:
```suggestion
        Default::default()
```

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    pub fn insert_extra_newline_at_last_newline(&mut self) {
+        let idx = self.index_of_prev_hard_newline();
+        let insert_idx = match idx {
+            Some(idx) => idx + 1,
+            None => 0,
+        };
+
+        self.contents.insert(insert_idx, LineToken::HardNewLine);
```
sgrif, line 50: I'm starting to think you want a `VecDeque` here. `insert` on a `Vec` is `O(n-i)`. On `VecDeque` it's `O(min(i, n-i))`. Same for `split_off` and pretty much any other function that moves elements at a random index.

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    pub fn last_token_is_a_newline(&self) -> bool {
+        self.contents
+            .last()
+            .map(|x| x.is_newline())
+            .unwrap_or(false)
```
sgrif, line 57:
```suggestion
            .map_or(false, |x| x.is_newline())
```

Clippy has lints for these

native/src/line_token_collection.rs
```diff
@@ -0,0 +1,83 @@
...
+            .map(|tok| tok.into_ruby().len())
+            .sum()
+    }
+
+    pub fn insert_comments_at_last_hard_newline(&mut self, cb: CommentBlock) {
+        let idx_of_prev_hard_newline = self.index_of_prev_hard_newline();
+        let insert_index = match idx_of_prev_hard_newline {
+            Some(idx) => idx + 1,
+            None => 0,
+        };
```
sgrif, line 79:
```suggestion
        let insert_index = self.index_of_prev_hard_newline().map_or(0, |i| i+1);
```

native/src/parser_state.rs
```diff
@@ -93,37 +85,33 @@ impl ParserState {
...
     }
 
     pub fn last_breakable_is_multiline(&self) -> bool {
-        self.breakable_entry_stack
-            .last()
-            .map(|o| o.is_multiline())
-            .unwrap_or(false)
+        self.target_stack
+            .current_breakable_is_multiline()
+            .expect("calling this without a breakable is a prgoramming error")
```
sgrif, line 106:
```suggestion
            .expect("calling this without a breakable is a programming error")
```

sgrif: Maybe have an error that reports this is a bug and links to the new issue page?

penelopezone:
@sgrif https://github.com/penelopezone/rubyfmt/pull/193

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
+use crate::breakable_entry::BreakableEntry;
+use crate::comment_block::CommentBlock;
+use crate::line_token_collection::LineTokenCollection;
+use crate::line_tokens::LineToken;
+use crate::types::LineNumber;
+
+#[derive(Debug)]
+pub enum RenderTarget {
+    BaseQueue(LineTokenCollection),
```
sgrif, line 9: What is a "base queue"?

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+            .insert_extra_newline_at_last_newline();
+    }
+
+    pub fn push_breakable_entry(&mut self, be: BreakableEntry) {
+        self.target_stack.push(RenderTarget::BreakableEntry(be));
+    }
+
+    pub fn pop_expecting_breakable_entry(&mut self) -> BreakableEntry {
+        if let RenderTarget::BreakableEntry(be) =
+            self.target_stack.pop().expect("should be present")
```
sgrif, line 78: Why?

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            return be;
+        }
+        panic!("should be impossible")
```
sgrif, line 82: Why?

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+        if self.target_stack.len() != 1 {
+            panic!("target stack was consumed when not length 1, indicates programming error");
+        }
+
+        let mut ts = self.target_stack;
+
+        match ts.pop().expect("we checked it exists") {
+            RenderTarget::BaseQueue(x) => x,
+            RenderTarget::BreakableEntry(_) => {
+                panic!("should not have breakable entry at bottom of stack")
```
sgrif, line 110: Is there any way we can use the type system to enforce these invariants? Should an issue be opened if they're violated?

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
```
sgrif, line 112:
```suggestion
        match self.target_stack.pop() {
            Some(RenderTarget::BaseQueue(x)) if self.target_stack.is_empty(),  => x,
            Some(RenderTarget::BreakableEntry(_)) => {
                panic!("should not have breakable entry at bottom of stack")
            }
            _ => panic!("target stack was consumed when not length 1, indicates programming error"),
        }
```

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    pub fn render_queue(&self) -> &LineTokenCollection {
+        match self.target_stack.get(0).expect("it's never empty") {
```
sgrif, line 116: Why?

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+            RenderTarget::BaseQueue(rq) => rq,
+            RenderTarget::BreakableEntry(..) => panic!("prgoramming error on get render queue"),
```
sgrif, line 118:
```suggestion
            RenderTarget::BreakableEntry(..) => panic!("programming error on get render queue"),
```

native/src/render_target_stack.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    fn current_target(&self) -> &RenderTarget {
+        self.target_stack
+            .last()
+            .expect("target stack is never empty")
```
sgrif, line 125: How about we make `current_target` a field of the struct then?

sgrif (commented):
I think this is over-using panics, I'd like to see more invariants enforced in the type system, or `Result`/`Option` used for fallible functions. Otherwise, this seems like a reasonable refactor.

PR #193, opened 2020-04-24
plz help

fables-tales:


sgrif:
Without a deeper change to how you're handling functions like these, your best bet is to use `RefCell` and have `with_breakable` take `&self` instead of `&mut self`

Issue #125, opened 2019-04-07
editor plugin for vscode


penelopezone:

andrewmcodes:
I’ll take a crack at it 👍🏼

samphippen:
Please!

Once it's complete open a pull request against this repo with a guide for how to install (imagine that I've never used vscode before).

I'll merge it down and add docs.

andrewmcodes:
I attempted this but ran into some problems and have not been able to come back to it for a few months. If someone else wants to give it a try please do; otherwise, I will try again when I have some time. In the meantime, I believe you can create after-save hooks without creating a plugin for vscode if anyone needs an intermediate solution.

wingrunr21:
Tracking issue:

rubyide/vscode-ruby#445

I will ship support for `v0.2.0` in the next release

wingrunr21:
Just an update here: `v0.25.0` included support

PR #201, opened 2020-06-03
Add instructions for setting up vscode integration

wadetandy:


wadetandy:
Fixes #125

penelopezone:
Hi @wadetandy I think unfortunately this doesn't work! If you look at `https://github.com/rubyide/vscode-ruby/blob/master/packages/language-server-ruby/src/formatters/RubyFMT.ts#L4` they vendor rubyfmt.rb (the legacy, broken) version of Rubyfmt in to their plugin. We should ask them either to update to point at the rust version or build our own plugin

wadetandy:
ahh interesting, didn't realize. I had run it against a few files in my project, but I guess the formatting issues were simple enough that this older bundled version worked fine. Opened an issue with the extension, so we can hopefully get the discrepancy resolved.

wadetandy:
@penelopezone Have opened [a PR](https://github.com/rubyide/vscode-ruby/pull/628) for vscode-ruby which the formatter to rely on a rubyfmt available on PATH.  As that doesn't work correctly with an alias, I've also updated this PR to include a script for generating a shim executable instead in place of a bash alias.  Glad to update it and/or the README however you'd like.

wadetandy:
Updated with a test for stdin processing, since I had separately started adding this for necessity with vscode integration and realized you'd already done it. Since I'd already written the test I figured I'd include it.

wadetandy:
Hey @penelopezone. Any feedback around why this was closed? Would love to figure out a path forward on supporting vscode and being able to leverage the existing plugin seems like the quickest path forward. Glad to rework this however you'd like, however.

fables-tales:
OH! This was closed because I deleted the master branch and rebased all the commits to remove my deadname. You should retarget main and be able to re-open it! I didn't mean to close this :) Sorry for the churn! This looks great, full review in a bit.

wadetandy:
Ahh makes sense. Reopened #204 to address.

PR #210, opened 2020-06-20
Even more ruby files

robotdana:
Reopened from #189

More files use ruby than just .rb. This is as exhaustive as i could make
it with a little bit of research (all the files i've come across as a
ruby developer, plus whatever documented variations there are). I definitely missed some https://github.com/rubocop-hq/rubocop/blob/c4ce01c333f962165576031e0ab8d61a4aafa8ed/config/default.yml#L11
Also checks ruby files with no extension, but with a ruby shebang

This is largely based on code from
https://github.com/robotdana/fast_ignore which in turn is based on code
from stdlib find but with fewer system calls so it's very slightly faster.

### DIFF
--- a/rubyfmt.rb
+++ b/rubyfmt.rb
@@ -210,24 +210,29 @@ def parse_file(fn)
   [file_data, Parser.new(file_data).parse]
 end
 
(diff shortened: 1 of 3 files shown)


fables-tales:
closing on the basis of very outdated

PR #212, opened 2020-06-20
Fix balanced string delims around interpolation

### LINKED ISSUE #211: %{{#{}}} crashes rubyfmt
I know that string looks horrible, but it a big graphql string tripped it and this is what i reduced it to. 

* Ruby version: 2.7.0 (also 2.6.5)
* Rubyfmt git sha: b2a96c54835b6f6f8801c9376127b20207a61ba9

## Input file

```ruby
puts %{{#{}}}
```

To break it there seems to need to be balanced parens of the same paren outside an interpolation.

other strings that break:
```ruby
%Q{a{#{}}}
%[[#{}]a]
%Q((#{}a))
```

## Rubyfmt's output:
```
Traceback (most recent call last):
        6: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:235:in `<main>'
        5: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:210:in `parse_file'
        4: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:210:in `parse'
        3: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:152:in `on_string_literal'
        2: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:152:in `each'
        1: from /Users/dana/Projects/rubyfmt/rubyfmt.rb:156:in `block in on_string_literal'
/Users/dana/Projects/rubyfmt/rubyfmt.rb:156:in `eval': (eval):1: unterminated string meets end of file (SyntaxError)
```

## This breaks because

It forces me to make my string less horrifying with my very own hands

robotdana:
Fixes: #211

The previous way of doing this with eval was very clever, but resulted
in trying to eval strings like `%{{}` which is just doesn't work. I
tried to think of ways we could continue to use this eval pattern
somehow but everything i could think of was more fragile than regular
expressions, so lets use regular expressions.

Because (I assume, please don't tell me this is incorrect) single quoted
and %q strings don't have interpolations we continue to use the original
eval pattern (i tried processing these with regular expressions as well
and they have completely different handling of escapes, so since this
works and is less fragile it can stay.

So we find the characters we need to escape after counting the
backslashes and either insert or remove a backslash as necessary.

Sorry i had to ruin strings again.

### DIFF
--- a/ci/string_literals_stress_test.rb
+++ b/ci/string_literals_stress_test.rb
@@ -1,3 +1,5 @@
+puts ''
+puts ""
 puts %Q("")
 puts %^_^
 puts %^\"^
@@ -12,7 +14,13 @@ puts '"'
 puts '\"'
 puts "\""
 puts "\\3\3"
+puts %^\\"\^^
+puts '\a^'
 puts %^\\"#{'\a^'}\^^
+puts %{{a#{1}}}
+puts %{{a}#{1}}
+puts %{\\\{#{1}}
+puts %{\\{#{1}}}
 puts <<EOD
 "abc"\"
 EOD
--- a/rubyfmt.rb
+++ b/rubyfmt.rb
@@ -147,21 +147,32 @@ class Parser < Ripper::SexpBuilderPP
       start_delim = @string_stack.pop
 
       if start_delim != "\""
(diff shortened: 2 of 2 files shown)


fables-tales:
This is amazing, thank you!!

PR #213, opened 2020-06-28
Embed ruby

fables-tales:
I am a monster


librubyfmt/build.rs
```diff
@@ -0,0 +1,39 @@
...
+            .arg(format!("{}/configure && make -j", ruby_checkout_path))
+            .current_dir(&ruby_checkout_path)
+            .output().expect("works1 ");
+        if !o.status.success() {
+            io::stdout().write_all(&o.stdout).unwrap();
+            io::stderr().write_all(&o.stderr).unwrap();
+            panic!("failed subcommand");
+        }
+    }
+    if !Path::new(&format!("{}/ruby_checkout/ruby-2.6.6/libripper.2.6-static.a", path.display())).exists() {
```
sgrif:
```suggestion
    if !path.join("ruby_checkout/ruby-2.6.6/libripper.2.6-static.a").exists() {
```

librubyfmt/build.rs
```diff
@@ -0,0 +1,39 @@
...
+            .current_dir(&ruby_checkout_path)
+            .output().expect("works");
+        if !o.status.success() {
+            panic!("failed subcommand");
+        }
+    }
+    cc::Build::new()
+        .file("src/rubyfmt.c")
+        .include(format!("{}/include", ruby_checkout_path))
+        .include(format!("{}/.ext/include/x86_64-darwin19", ruby_checkout_path))
```
sgrif: "Works on my machine"

librubyfmt/build.rs
```diff
@@ -0,0 +1,39 @@
+use std::process::Command;
+use std::io::{self, Write};
+use std::path::Path;
+
+fn main() {
+    let path = std::env::current_dir().expect("is current");
+    let ruby_checkout_path = format!("{}/ruby_checkout/ruby-2.6.6", path.display());
+    if !Path::new(&format!("{}/libruby.2.6-static.a", ruby_checkout_path)).exists() {
```
sgrif:
```suggestion
    if !path.join("ruby_checkout/ruby-2.6.6/libruby.2.6-static.a").exists() {
```

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+impl FormatBuffer {
+    pub fn into_buf(self) -> &'static [u8] {
+        unsafe { slice::from_raw_parts(self.bytes as *const u8, self.count as usize) }
+    }
+
+    pub fn into_string(self) -> String {
+        unsafe {
+            let vec = Vec::from_raw_parts(self.bytes as *mut u8, self.count as usize, self.count as usize);
+            let res = String::from_utf8_unchecked(vec);
+            return res
```
sgrif:
```suggestion
            String::from_utf8_unchecked(vec)
```

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+extern "C" {
+    pub fn Init_ripper();
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct FormatBuffer {
+    pub bytes: *const libc::c_char,
+    pub count: i64,
+}
```
sgrif: What do you think about having this wrap `Box<str>` or `Box<CStr>`?

fables-tales: why?

sgrif: Because you're reimplementing it anyway, why not do it without unsafe code

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[no_mangle]
+pub extern "C" fn rubyfmt_format_buffer(buf: FormatBuffer) -> FormatBuffer {
+    let output_data = Vec::with_capacity(buf.count as usize);
+    let mut output = Cursor::new(output_data);
+    let tree = run_parser_on(buf);
+    if tree.is_err() {
+        unsafe {
+            let cstr = CString::new("oh no").expect("we just made it");
+            ruby::rb_raise(ruby::rb_eRuntimeError, cstr.as_ptr());
```
sgrif:
```suggestion
            ruby::rb_raise(ruby::rb_eRuntimeError, "oh no\0".as_ptr() as _);
```

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+    let tree = tree.expect("we raised");
```
sgrif:
```suggestion
    let tree = match run_parser_on(buf) {
        Err(_) => return ruby::rb_raise(ruby::rb_eRuntimeError, "oh no\0".as_ptr() as _),
        Ok(tree) => tree,
    }
```

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+    raise_if_error(res);
+    let output_data = output.into_inner().into_boxed_slice();
+    let ptr = output_data.as_ptr();
+    let len = output_data.len();
+    let fb = FormatBuffer {
+        bytes: ptr as *const libc::c_char,
+        count: len as i64,
+    };
+    std::mem::forget(output_data);
+    fb
```
sgrif: I know you're getting rid of `FormatBuffer`, but this is how I'd write this if you were keeping it. Also worth noting that this leaks, and I'd consider including the capacity on the struct to avoid the realloc here.

```suggestion
    output.get_mut().shrink_to_fit();
    let (ptr, len, _) = output.into_inner().into_raw_parts();
    FormatBuffer {
        bytes: ptr as *const libc::c_char,
        count: len as i64,
    }
```

src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+        eprintln!("4");
+        io::stdout().flush().expect("flush works");
+    } else if args[1] == "-i" {
+        let parts = &args[2..args.len()];
+        format_parts(parts);
+    } else {
+        let parts = &args[1..args.len()];
+        format_parts(parts);
+    }
+}
```
sgrif: It sounds like the point of this is to test that the API C would call actually works, but it's worth noting that you can skip a lot of allocation and buffering by calling into `toplevel_format_program`, passing `stdout` or the file that you're writing to directly

src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+use glob::glob;
+
+fn rubyfmt_file(file_path: PathBuf) -> io::Result<()> {
+    let buffer = read_to_string(file_path.clone())?;
+    let res = rubyfmt::format_buffer(buffer);
+    let mut file = OpenOptions::new()
+        .write(true)
+        .open(file_path)
+        .expect("file");
+    write!(file, "{}", res)?;
```
sgrif: Same thing here, I'd consider having `format_buffer` take `&mut Write` and pass the file directly (probably with a `BufWriter`)

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+        "../ruby_checkout/ruby-2.6.6/ext/ripper/lib/ripper/filter.rb"
+    ))?;
+    eval_str(include_str!(
+        "../ruby_checkout/ruby-2.6.6/ext/ripper/lib/ripper/sexp.rb"
+    ))?;
+
+    Ok(())
+}
+
+fn eval_str(s: &str) -> Result<VALUE, ()> {
```
sgrif: One thing I noticed is that you never call this with anything other than `&'static str`. If the input string were always nul terminated you could skip some copies (but that would require either taking `CStr`, an annoying type to construct, or assuming `&str` is always nul terminated, which would make it unsafe). However, since you are only ever taking literals, you could make this a macro and safely do it.

```rust
macro_rules! eval_str {
    ($s:expr) => {
        let mut state = 0;
        ruby::rb_eval_string_protect(
            concat!(s, "\0").as_ptr(),
            &mut state,
        );
        if state != 0 {
            return Err(())
        }
    }
}
```

(note: I omitted the return value since it's never used)

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+    InitStatus::OK as libc::c_int
+}
+
+pub fn format_buffer(buf: String) -> String {
+    eprintln!("format 1");
+    let bytes: Vec<libc::c_char> = buf
+        .into_bytes()
+        .into_iter()
+        .map(|v| v as libc::c_char)
+        .collect();
```
sgrif: This is moot since you're getting rid of `FormatBuffer`, but you could just cast the pointer for the same effect but without a copy

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+        unsafe {
+            let vec = Vec::from_raw_parts(
+                self.bytes as *mut u8,
+                self.count as usize,
+                self.count as usize,
+            );
+            String::from_utf8_unchecked(vec)
+        }
+    }
+}
```
sgrif: So *if* you were to try to keep this type, here's how I would write it to eliminate the unsoundness. I've noted the differences where they occur

```rust
#[repr(C)]
#[derive(Debug)] // This represents an owned, heap allocated pointer. To ensure it doesn't leak, it must have a `Drop` impl. It may also have methods that return similar owned types like `Vec`, `String`, or `Box`. It absolutely should not impl `Copy`, and a `Clone` impl would need to be manual and copy the memory being pointed to.

struct FormatBuffer {
    ptr: NonNull<u8>, // Encode as much as we can here. Using `u8`, C can cast the pointer all they want.
    len: usize, // Again, using the Rust type that's convenient. This is size_t or uintptr_t in C.
}

impl FormatBuffer {
    // Taking Into<Vec<u8>> since you're calling with String and Vec<u8>. `into_boxed_slice` copies, we can avoid
    // that if we keep the capacity around as well. Most importantly though, we provide a constructor which consumes
    // the input, leaving now room for one caller to accidentally use `.as_ptr()` instead of into_raw or similar. This 
    // struct should also live in its own file so its fields are private (C can of course see them, but we Rust callers
    // shouldn't).
    fn new<T: Into<Vec<u8>>>(bytes: T) -> Self {
        let bytes = bytes.into().into_boxed_slice();
        let len = bytes.len();
        Self {
            // safety: Box::into_raw never returns null. Use Box::into_raw_nonnull when stable
            ptr: unsafe { NonNull::new_unchecked(Box::into_raw(bytes) as _) },
            len,
        }
    }

    // Does not return 'static. Instead returns a proper lifetime.
    pub fn as_bytes(&self) -> &[u8] {
        // safety: We know we were constructed with a valid pointer
        unsafe { slice::from_raw_parts(self.bytes, self.len) }
    }

    // Returns `&str` instead of `String`. We own the pointer, we don't want to be returning
    // other types which own it.
    pub fn as_str(&self) -> &str {
        // Uses checked, we didn't validate UTF-8 on input. Even if we did, let's avoid unsafe
        str::from_utf8(self.as_bytes()).expect("Invalid UTF-8")
    }
}

impl Drop for FormatBuffer {
    fn drop(&mut self) {
        unsafe {
            let ptr = ptr::slice_from_raw_parts(self.ptr, self.len);
            drop(Box::from_raw(ptr));
        }
    }
}
```

librubyfmt/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+
+fn run_parser_on(buf: FormatBuffer) -> Result<VALUE, ()> {
+    unsafe {
+        let buffer_string = ruby::rb_utf8_str_new(buf.bytes, buf.count);
+        let parser_class = eval_str("Parser")?;
+        let parser_instance = ruby::rb_funcall(parser_class, intern("new"), 1, buffer_string);
+        let tree = ruby::rb_funcall(parser_instance, intern("parse"), 0);
+        Ok(tree)
+    }
+}
```
sgrif:
```suggestion
macro_rules! intern {
    ($s:str) => {
        unsafe { ruby::rb_intern(concat!($s, "\0".as_ptr*() as _)) }
    }
}

fn run_parser_on(buf: FormatBuffer) -> Result<VALUE, ()> {
    unsafe {
        let buffer_string = ruby::rb_utf8_str_new(buf.bytes, buf.count);
        let parser_class = eval_str("Parser")?;
        let parser_instance = ruby::rb_funcall(parser_class, intern!("new"), 1, buffer_string);
        let tree = ruby::rb_funcall(parser_instance, intern!("parse"), 0);
        Ok(tree)
    }
}
```

fables-tales: did this

PR #222, opened 2020-07-12
improve workflow script code style

chenrui333:
Improve github action workflow code style


fables-tales:
This is helpful! Thank you so much :)

PR #226, opened 2020-08-06
Fix unescaped square brackets and escaped delims in percent arrays

### LINKED ISSUE #225: `%w{...]...}` is wrongly formatted to `%w[...]...]`
* Ruby version: ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
* Rubyfmt git sha: 3f3653ccc6210a37eab2954ad534c901d63604a4

## Input file

```ruby
%w{...]...}
```

## Rubyfmt's output

```
%w[...]...]
```

## This breaks because

The delimiters of `%w{}` is changed to brackets, and the contents include a closed bracket which unintentionally terminates the `%w` literal. I have no good idea how to fix, though.

robotdana:
e.g. `%w{[ ]}` => `%w[\[ \]]` (which is equivalent to `["[", "]"]`)

fixes: #225

This does mean that this is possibly unnecessarily escaped, because
balanced pairs are fine in percent literals BUT this is easier, and
slightly more explicit.

i also added a hacky way to transform `%w{\{}`

before it would change the value from `['{']` to `['\\{']`. now it doesn't. 
but christ is it a hack

(also i'm unclear which ruby file i'm supposed to be modifying so i
modified them all)

**ALTERNATIVELY TO THIS PR**, we push the delim to the ast and don't change it
to another delim (might be cleaner in the end, but that sounds like a rust task and my
rust is weak at best)

**ALTERNATIVELY ALTERNATIVELY**, we transform things even more to regular array & string/symbol literals

### DIFF
--- a/script/tests/test_array_literals.sh
+++ b/script/tests/test_array_literals.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -ex
+
+source ./script/functions.sh
+
+ARRAY_LITERALS_EXPECTED=$(ruby ci/array_literals_stress_test.rb | f_md5)
+ARRAY_LITERALS_ACTUAL=$(f_rubyfmt ci/array_literals_stress_test.rb | ruby | f_md5)
+if [[ "$ARRAY_LITERALS_EXPECTED" != "$ARRAY_LITERALS_ACTUAL" ]]
+then
+    echo "array literals are broken"
+    exit 1
+fi
--- a/ci/array_literals_stress_test.rb
+++ b/ci/array_literals_stress_test.rb
@@ -29,5 +29,46 @@ puts %W^[#{1}]^.inspect
 puts %W^[]#{1}^.inspect
 puts %W^\[^.inspect
 puts %W^\^^.inspect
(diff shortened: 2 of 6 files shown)


robotdana:
(not clear why the build failed, it built fine on my mac)

fables-tales:
@robotdana this failure is 100% unrelated to your change, I will fix it when I can (I'm recovering from surgery right now!)

robotdana:
❤️

script/tests/test_array_literals.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
fables-tales, line 1: @robotdana you need to add this to https://github.com/penelopezone/rubyfmt/blob/trunk/script/test.sh for it to run in CI. it looks like it's missing right now?

fables-tales (changes requested):

fables-tales:
manually merged in da5c3631779121f5281726c093f0548e00cf8034

PR #230, opened 2020-08-23
Refactor ruby interface

fables-tales:


librubyfmt/src/comment_block.rs
```diff
@@ -13,10 +14,25 @@ impl CommentBlock {
...
 
+    pub fn apply_spaces(self, indent_depth: ColNumber) -> Self {
+        let new_strings = self
+            .comments
+            .into_iter()
+            .map(|c| {
+                let spaces = (0..indent_depth)
+                    .map(|_| " ".to_string())
+                    .collect::<Vec<String>>()
+                    .join("");
```
sgrif:
```suggestion
                let spaces = str::repeat(" ", indent_depth)
```

librubyfmt/src/ruby_ops.rs
```diff
@@ -0,0 +1,82 @@
...
+            }
+        } else {
+            let s = current_exception_as_rust_string();
+            Err(ParseError::OtherRubyError(s))
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct RipperTree(VALUE);
```
sgrif, line 72: Given that this just has a safe constructor and never asserts that the inner type is actually what is expected, I'm not entirely sure what value this brings

librubyfmt/src/file_comments.rs
```diff
@@ -1,51 +1,54 @@
...
 
-                let should_push = line_number == 1
-                    || (last_line.is_some() && last_line.unwrap() == &(line_number - 1));
-                if should_push {
-                    res.contiguous_starting_indices.push(line_number);
-                }
-                res.comment_blocks.insert(line_number, l);
-            }
+        let should_push =
+            line_number == 1 || (last_line.is_some() && last_line.unwrap() == &(line_number - 1));
```
sgrif:
```suggestion
        let should_push = last_line.copied().unwrap_or(0) == line_number - 1
```

librubyfmt/src/file_comments.rs
```diff
@@ -1,51 +1,54 @@
...
+        if should_push {
+            self.contiguous_starting_indices.push(line_number);
```
sgrif: I need to give a higher level overview of what this is for, but my gut tells me this is sidestepping the abstraction we need or that what this is could be made more clear

librubyfmt/src/file_comments.rs
sgrif: Or if my assumption that `last_line` is always `None` when `line_number == 1` is incorrect,
```suggestion
        let should_push =
            line_number == 1 || last_line.copied() == Some(line_number - 1);
```

librubyfmt/src/file_comments.rs
```diff
@@ -59,7 +62,11 @@ impl FileComments {
 
         let mut sled = Vec::with_capacity(self.contiguous_starting_indices.len());
         for key in self.contiguous_starting_indices.iter() {
-            sled.push(self.comment_blocks.remove(key).expect("we tracked it"));
+            sled.push(
+                self.comment_blocks
+                    .remove(key)
+                    .unwrap_or_else(|| panic!("we tracked it: {} {:?}", key, self.comment_blocks)),
```
sgrif: This panic is also making me think there's a better structure for this

librubyfmt/src/lib.rs
```diff
@@ -76,28 +76,30 @@ pub enum FormatError {
...
 }
 
 #[no_mangle]
 pub extern "C" fn rubyfmt_init() -> libc::c_int {
     init_logger();
-    unsafe {
-        ruby::ruby_init();
+    let res = ruby_ops::setup_ruby();
+    if res.is_err() {
+        return InitStatus::ERROR as libc::c_int;
```
sgrif, line 93: Should the C API provide a way to get a human readable description of the error?

```rust
static mut LAST_ERROR: Option<CString> = None;

// on this line
if let Err(e) = res {
    // If we think there could be multi-threading here we should use `lazy_static!` and a mutex
    unsafe { LAST_ERROR = Some(CString::new(e.to_string()).expect("don't actually expect in prod")) }
    return InitStatus::ERROR as libc::c_int;
}

// later
#[no_mangle]
pub extern "C" fn rubyfmt_last_error -> *const libc::c_char {
    match LAST_ERROR {
        Some(e) => e.as_ptr(),
        None => ptr::null(),
    }
}
```

librubyfmt/src/parser_state.rs
```diff
@@ -320,22 +312,21 @@ impl ParserState {
                 None => 0,
             };
 
+            let spaces = self.spaces_after_last_newline;
+            debug!("spaces: {} comments: {:?}", spaces, self.comments_to_insert);
+            let bt = Backtrace::new();
+            debug!("{:?}", bt);
```
sgrif: If we're at the point where we're printing backtraces into the debug logger, we should probably use the `tracing` library instead

librubyfmt/src/parser_state.rs
```diff
@@ -487,6 +494,9 @@ impl ParserState {
         // valid utf8 and also we're only using this to newline match
         // which should be very hard to break. The unsafe conversion
         // here skips a utf8 check which is faster.
+        // FIXME: Is the overhead of utf8 checking actually a bottleneck here?
+        // The comment above even admits there are circumstances where it will
+        // not be UTF-8
         unsafe {
             let s = str::from_utf8_unchecked(&data).to_string();
             s.trim().chars().any(|v| v == '\n')
```
sgrif:
```suggestion
            }
        s.trim().contains('\n')
```

(It won't let me include the closing brace that I'm moving in this suggestion comment sorry

librubyfmt/src/parser_state.rs
```diff
@@ -503,10 +513,7 @@ impl ParserState {
     pub fn render_heredocs(&mut self, skip: bool) {
         while !self.heredoc_strings.is_empty() {
             let mut next_heredoc = self.heredoc_strings.pop().expect("we checked it's there");
-            let want_newline = match self.render_queue.last() {
-                Some(x) => !x.is_newline(),
-                None => true,
-            };
+            let want_newline = !self.current_target_mut().last_token_is_a_newline();
```
sgrif:
```suggestion
            let want_newline = !self.current_target().last_token_is_a_newline();
```

librubyfmt/src/parser_state.rs
```diff
@@ -624,4 +624,24 @@ impl ParserState {
...
     }
+
+    fn current_target(&self) -> &dyn LineTokenTarget {
+        if self.breakable_entry_stack.is_empty() {
+            &self.render_queue
+        } else {
+            self.breakable_entry_stack
+                .last()
+                .expect("we checked it's not empty")
+        }
```
sgrif, line 618:
```suggestion
        self.breakable_entry_stack
            .last()
            .unwrap_or(&self.render_queue)
```

librubyfmt/src/parser_state.rs
```diff
@@ -624,4 +624,24 @@ impl ParserState {
...
+    }
+
+    fn current_target_mut(&mut self) -> &mut dyn LineTokenTarget {
+        if self.breakable_entry_stack.is_empty() {
+            &mut self.render_queue
+        } else {
+            self.breakable_entry_stack
+                .last_mut()
+                .expect("we checked it's not empty")
+        }
```
sgrif:
```suggestion
        self.breakable_entry_stack
            .last_mut()
            .unwrap_or(&mut self.render_queue)
```

librubyfmt/src/render_targets.rs
```diff
@@ -3,11 +3,58 @@ use crate::line_tokens::LineToken;
 use crate::types::{ColNumber, LineNumber};
 use std::collections::HashSet;
 
+fn insert_at<T>(idx: usize, target: &mut Vec<T>, input: &mut Vec<T>) {
+    let drain = input.drain(..);
+    let mut idx = idx;
+    for item in drain {
+        target.insert(idx, item);
+        idx += 1;
+    }
```
sgrif:
```suggestion
    let mut tail = target.split_off(idx);
    input.append(input);
    input.append(&mut tail);
```

sgrif: The code as written will perform as many as log(N) allocations and 2N memcpys. The new code will perform at most 2 and at minimum 1 allocations, and 2 memcpys

librubyfmt/src/render_targets.rs
```diff
@@ -3,11 +3,58 @@ use crate::line_tokens::LineToken;
...
+pub struct BaseQueue {
+    tokens: Vec<LineToken>,
+}
+
+impl LineTokenTarget for BaseQueue {
+    fn push(&mut self, lt: LineToken) {
+        self.tokens.push(lt)
+    }
+
+    fn insert_at(&mut self, idx: usize, tokens: &mut Vec<LineToken>) {
```
sgrif, line 36: It looks like this is mostly called with a single element, should this just take that element instead of a vec?

librubyfmt/src/render_targets.rs
```diff
@@ -3,11 +3,58 @@ use crate::line_tokens::LineToken;
...
     SingleLine,
 }
 
+pub trait LineTokenTarget {
+    fn push(&mut self, lt: LineToken);
+    fn insert_at(&mut self, idx: usize, tokens: &mut Vec<LineToken>);
+    fn into_tokens(self, ct: ConvertType) -> Vec<LineToken>;
+    fn last_token_is_a_newline(&self) -> bool;
+    fn index_of_prev_hard_newline(&self) -> Option<usize>;
+}
```
sgrif, line 24: The impl of every method on this trait is exactly the same (or could be). How about this instead?
```suggestion
pub trait LineTokenTarget {
    fn tokens(&self) -> &[LineToken];
    fn tokens_mut(&mut self) -> &mut Vec<LineToken>;
    
    fn push(&mut self, lt: LineToken) {
        self.tokens_mut().push(lt);
    }
    
    // etc
}
```

librubyfmt/src/render_targets.rs
```diff
@@ -47,6 +89,30 @@ impl BreakableEntry {
...
+            .iter()
+            .rposition(|v| v.is_newline() || v.is_comment())
+    }
+}
+
+impl BreakableEntry {
+    pub fn new(spaces: ColNumber, delims: BreakableDelims) -> Self {
+        BreakableEntry {
+            spaces,
+            tokens: vec![],
```
sgrif: Seeing `vec![]` next to `HashSet::new()` feels weird to me

```suggestion
            tokens: Vec::new(),
```

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
+        method_id: ID,
+        argc: libc::c_int,
+        argv: *const VALUE,
+        block: extern "C" fn(_: VALUE, _: VALUE, _: libc::c_int, _: *const VALUE) -> VALUE,
+        outer_scope: VALUE,
+    ) -> VALUE;
 }
 
 pub fn current_exception_as_rust_string() -> String {
+    let ruby_string = unsafe { eval_str("$!.inspect") }.expect("Error evaluating `$!.inspect`");
```
sgrif, line 97: Rather than panicking, how about we return a string indicating that there was an error even trying to get the value out of Ruby. Or even better, is there a function in the Ruby C API that we can call instead of evaling?

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
+
+macro_rules! intern {
+    ($s:literal) => {
+        rb_intern(concat!($s, "\0").as_ptr() as _)
+    };
+}
+
+// Safety: This function expects an initialized Ruby VM capable of evaling code
+pub unsafe fn eval_str(s: &str) -> Result<VALUE, ()> {
+    let rubyfmt_program_as_c = CString::new(s).expect("unexpected nul byte in Ruby code");
```
sgrif, line 109: Can we return an error instead of panicking?

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
+    }
+}
+
+extern "C" fn real_debug_inspect(v: VALUE) -> VALUE {
     unsafe {
-        let res = eval_str("$!.inspect").expect("this can't fail");
-        let ptr = rubyfmt_rstring_ptr(res);
-        let length = rubyfmt_rstring_len(res);
-        String::from_raw_parts(ptr as _, length as _, length as _)
+        let inspect = rb_funcall(v, intern!("inspect"), 0, std::ptr::null() as *const VALUE);
```
sgrif:
```suggestion
        let inspect = rb_funcall(v, intern!("inspect"), 0, std::ptr::null());
```

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
+        let char_pointer = rb_string_value_cstr(&inspect) as *mut i8;
+        let cstr = CStr::from_ptr(char_pointer);
+        let s = cstr.to_str().expect("it's utf8");
+        debug!("{}", s);
```
sgrif:
```suggestion
        debug!("{}", unsafe { ruby_string_to_str(inspect) });
```

librubyfmt/src/ruby.rs
```diff
@@ -50,13 +51,13 @@ pub const Qnil: VALUE = VALUE(8);
...
     // stuff that we need to compile out rubyfmt
-    pub fn ruby_init();
+    pub fn ruby_setup() -> libc::c_int;
     pub fn ruby_cleanup(_: libc::c_int);
     pub fn rb_eval_string_protect(_: *const libc::c_char, _: *mut libc::c_int) -> VALUE;
     pub fn rb_funcall(_: VALUE, _: ID, _: libc::c_int, ...) -> VALUE;
     pub fn rb_utf8_str_new(_: *const libc::c_char, _: libc::c_long) -> VALUE;
     pub fn rb_str_new_cstr(_: *const libc::c_char) -> VALUE;
-    pub fn rb_string_value_cstr(_: VALUE) -> *const libc::c_char;
+    pub fn rb_string_value_cstr(_: *const VALUE) -> *const libc::c_char;
```
sgrif, line 60: I don't see a corresponding change to the definition. Are you sure this is correct?

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
-            rubyfmt_program_as_c.as_ptr(),
-            &mut state as *mut libc::c_int,
-        );
+        rb_protect(real_debug_inspect as _, v, &mut state);
         if state != 0 {
-            Err(())
-        } else {
-            Ok(v)
+            let s = current_exception_as_rust_string();
+            panic!("blew us: {}", s);
```
sgrif, line 133: Are you sure panicking if we failed to perform a debug inspection is the right call?

librubyfmt/src/ruby.rs
```diff
@@ -81,29 +83,82 @@ extern "C" {
...
         }
     }
 }
+
+pub fn raise(s: &str) {
+    let cstr = CString::new(s).expect("it's not null");
```
sgrif: "it's not null" doesn't accurately describe this error condition. Can we just handle a nul byte being contained in the string gracefully or at least give a useful error message?

sgrif (approved):
I don't see any obvious UB. I would still like to improve the structure of a lot of this and/or attempt to make it safe in the future. I also had a few comments about style. I really think you should try to reduce the number of places this code base uses panics. But overall I don't see anything too heinous

librubyfmt/src/parser_state.rs
fables-tales: manually applied

librubyfmt/src/render_targets.rs
fables-tales: applied manually

librubyfmt/src/lib.rs
fables-tales: the header defines this: https://github.com/penelopezone/rubyfmt/blob/trunk/librubyfmt/include/rubyfmt.h#L4

librubyfmt/src/parser_state.rs
fables-tales: this was pretty one-off. I'll probably just delete

librubyfmt/src/ruby.rs
fables-tales: https://github.com/ruby/ruby/blob/7bafc9f433223d82aac8e8974b9bfdbdc34ca4df/include/ruby/internal/core/rstring.h#L92

PR #232, opened 2020-08-24
Partial audit of unsafe code, refactor to reduce its usage

sgrif:
This is the result of a partial audit of all unsafe code in rubyfmt. I
found one double free, and many instances where unsafe was used more
than it was needed. In some cases this was caused by omitting extremely
cheap safety checks like UTF-8 validation, and in some cases it was
redundant code converting arrays and strings to Rust versions. I've
added UTF-8 checking where it's clearly not going to cause a performance
problem, commented where it's not clear, and moved array and string
conversion to a common function that's easier to audit.

### DIFF
--- a/librubyfmt/src/ruby.rs
+++ b/librubyfmt/src/ruby.rs
@@ -1,6 +1,6 @@
 #![allow(non_camel_case_types, dead_code)]
 use log::debug;
-use std::ffi::{CString, CStr};
+use std::ffi::{CStr, CString};
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[repr(transparent)]
@@ -94,10 +94,8 @@ extern "C" {
 }
 
 pub fn current_exception_as_rust_string() -> String {
-    let ruby_string = unsafe { eval_str("$!.inspect") }
-        .expect("Error evaluating `$!.inspect`");
-    unsafe { ruby_string_to_str(ruby_string) }
-        .to_owned()
+    let ruby_string = unsafe { eval_str("$!.inspect") }.expect("Error evaluating `$!.inspect`");
+    unsafe { ruby_string_to_str(ruby_string) }.to_owned()
 }
 
 macro_rules! intern {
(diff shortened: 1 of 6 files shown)


librubyfmt/src/ruby.rs
```diff
@@ -93,12 +94,10 @@ extern "C" {
 }
 
 pub fn current_exception_as_rust_string() -> String {
-    unsafe {
-        let res = eval_str("$!.inspect").expect("this can't fail");
-        let ptr = rubyfmt_rstring_ptr(res);
-        let length = rubyfmt_rstring_len(res);
-        String::from_raw_parts(ptr as _, length as _, length as _)
```
sgrif, line 100: This was the double free mentioned in the PR description. Clearly this branch isn't executed on any files we currently test on, but this would result in the same segfault seen earlier

librubyfmt/src/ruby.rs
```diff
@@ -151,3 +144,21 @@ pub fn raise(s: &str) {
         rb_raise(rb_eRuntimeError, cstr.as_ptr());
     }
 }
+
+// Safety: The given VALUE must be a valid Ruby array. The lifetime is not
+// checked by the compiler. The returned slice must not outlive the given
+// Ruby array. The Ruby array must not be modified while the returned slice
+// is live.
+pub unsafe fn ruby_array_to_slice<'a>(ary: VALUE) -> &'a [VALUE] {
+    std::slice::from_raw_parts(rubyfmt_rb_ary_ptr(ary), rubyfmt_rb_ary_len(ary) as _)
```
fables-tales, line 153: If we're building abstractions around this, let's make them safe(r). If you use `rubyfmt_rb_type` here to check if the type is `T_ARRAY` and return a result, that would be cleaner right?

fables-tales: or if we're relying on this being called from `rb_protect`, can we document this will ruby raise and segfault if not inside an rb protect call?

librubyfmt/src/ruby.rs
```diff
@@ -151,3 +144,21 @@ pub fn raise(s: &str) {
...
+}
+
+// Safety: The given VALUE must be a valid Ruby string. The lifetime is not
+// checked by the compiler. The returned str must not outlive the given
+// Ruby string. The Ruby string must not be modified while the returned str
+// is live.
+pub unsafe fn ruby_string_to_str<'a>(s: VALUE) -> &'a str {
```
fables-tales, line 160: same thing here.

fables-tales (changes requested):

librubyfmt/src/ruby.rs
sgrif: Yes, that's an additional step I'd like to take *if* we can guarantee that `VALUE` is always a valid pointer, but if we're requiring the caller to verify invariants anyway we may as well allow this one as well

fables-tales (approved):

PR #235, opened 2020-08-29
Refactor `FileComments` to better demonstrate intent

sgrif:
I'm not particularly happy with how this came out, but I do think it's a
clear improvement over what was there before. Previously this has three
obviously interacting fields, with a lot of subtle, undocumented,
unmaintained invariants (or at least in the case of `lowest_key` it was
unmaintained). I do believe that this structure at least better
represents intent.

`CommentBlock` now keeps track of what lines it's on, and the "start of
file sled" (what even is that?!) is kept separately from the rest of the
comments. This reduces (but does not eliminates) the amount of state
that this type has to keep, and makes the intent clearer both in the
type itself, and its state at any single point. At absolute minimum, I
have at least removed all panics from this type's implementation.

That said, this interface has a ton of pitfalls. I really dislike how
much it needs to mutate itself. While developing this patch, it became
clear to me why this is the case. The parser that we've written only
retains the line number and the comment on that line. This means we have
no way to determine if two comments are part of a contiguous block or
not, and it seems we leave the formatter to determine whether that's the
case. We do treat the first comment block specially though, and assume
it's a contiguous block (though I'm not entirely clear why yet).

This is a problem for a number of reasons. First of all, it means that
we will just delete comments if there are more than one on a given line,
potentially breaking comments on method arguments that are significant
to rdoc.

This also misbehaves if the first non-comment line has a trailing
comment (most likely for rdoc, but could also potentially happen if the
first comment is for the file itself, and the trailing comment is a
poorly placed, very short comment for the item it's on)

This type very clearly wants to be `Map<Range<LineNumber>, Vec<String>>`
or even better `Vec<Span, String>` since the formatter should be able to
choose/change where newlines are inserted, and really shouldn't be
coupled to line numbers in the source. I'm sure Ripper *must* give us a
way to determine if non-comment tokens appear between two comments, but
I do not know enough about that layer of the code to fix it in this
patch. The fact that `FileComments` is constructed outside of the main
deserializer is surprising, and I would hope that we can perhaps at
least handle this in a `Deserialize` impl`.

But short of what I'd consider to be a more complete fix, I have at
least tried to make the invariants of this type better upheld, and those
that aren't enforced are now documented (mostly in `push_comment`, which
is still extremely subtle)

I don't think `push_comment` is the right interface for this type.
However, since it was marked as public, I've assumed that it is an
interface intended to be externally consumed and thus needs to be
maintained. If that's not the case, I think we should replace it with
something that only operates when we have a full view of the data, such
as `FromIterator`. In that case we can either manually sort the data
before consuming it, or at least include a debug assertion that it's
already sorted.

Speaking of debug assertions, one thing that stood out to me while
writing this is that the test suite is run in release mode. I'd like to
learn more about why that is.  Previous iterations of this patch
included debug assertions for the invariants that were implied but not
upheld. While none of these landed in the final patch, if they did I
feel like they should be run as part of the test suite.

### DIFF
--- a/librubyfmt/src/file_comments.rs
+++ b/librubyfmt/src/file_comments.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::mem;
 
 use crate::comment_block::CommentBlock;
 use crate::ruby::*;
@@ -6,9 +7,8 @@ use crate::types::LineNumber;
 
 #[derive(Debug, Default)]
 pub struct FileComments {
-    comment_blocks: BTreeMap<LineNumber, String>,
-    contiguous_starting_indices: Vec<LineNumber>,
-    lowest_key: LineNumber,
+    start_of_file_sled: Option<CommentBlock>,
+    other_comments: BTreeMap<LineNumber, String>,
 }
 
 impl FileComments {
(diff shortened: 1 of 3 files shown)


librubyfmt/src/file_comments.rs
```diff
@@ -36,58 +36,48 @@ impl FileComments {
...
+            (None, 1) => {
+                debug_assert!(
+                    self.other_comments.is_empty(),
+                    "If we have a start of file sled, it needs to come first,
+                     otherwise we won't know where the last line is",
+                );
+                self.start_of_file_sled = Some(CommentBlock::new(1..2, vec![l]));
+            }
+            (Some(sled), _) if sled.following_line_number() == line_number => {
+                sled.add_line(l);
```
sgrif, line 53: `l` looks like `1` here :\

sgrif: (I would pick a different variable name but I don't want to pointlessly churn the line declaring it. GH plz pick a less bad font)

librubyfmt/src/parser_state.rs
```diff
@@ -158,8 +157,7 @@ impl ParserState {
     }
 
     pub fn insert_comment_collection(&mut self, comments: CommentBlock) {
-        self.comments_to_insert
-            .merge(comments.apply_spaces(self.spaces_after_last_newline));
+        self.comments_to_insert += comments.apply_spaces(self.spaces_after_last_newline);
```
fables-tales, line 160: I am not sure if I hate this or if I love this. I will think about it

sgrif: It was just the most convenient way to get it onto `Option` but we can do a trait with a named method

PR #243, opened 2020-09-07
Fix bug with yield arguments within ifop

### LINKED ISSUE #227: Issue with yield self within a ternary operator
After your NoRuKo talk I tested against our codebase and this issue was immediately obvious.

* Ruby version: ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin19]
* Rubyfmt git sha: 3435618

## Input file

```ruby
def foo(&block)
  block_given? ? yield(self) : self
end
```

## Rubyfmt's output

```
def foo(&block)
  block_given? ? yield self : self
end
```

## This breaks because

Output is syntactically incorrect, Ruby errors with `syntax error, unexpected self, expecting ':'`.

kaoru:
Fixes https://github.com/penelopezone/rubyfmt/issues/227

-------------------

**Fix bug with yield arguments within ifop**

A call to yield with arguments within an IfOp (which is what the ruby parser calls a ternery hook) would remove the parentheses from the yield call, which would then produce code that Ruby could not parse because of the precedence of the `?` and `:` operators.

This commit fixes that by ensuring that within an IfOp context we always keep the parentheses around a yield call if it has arguments.

--------------------

Thanks for pointing me in the right direction @penelopezone, your instructions with file names and line numbers were super helpful! 

My commit only adds tests for `yield` but the change would also affect `return`, `raise` and `break` in an IfOp context. I think that's correct because those also fail to parse:

```
$ ruby -c -e 'x ? yield x : nil' 
-e:1: syntax error, unexpected tIDENTIFIER, expecting ':'
x ? yield x : nil

$ ruby -c -e 'x ? return x : nil'
-e:1: syntax error, unexpected tIDENTIFIER, expecting ':'
x ? return x : nil

$ ruby -c -e 'x ? break x : nil' 
-e:1: syntax error, unexpected tIDENTIFIER, expecting ':'
x ? break x : nil

$ ruby -c -e 'x ? raise x : nil'
-e:1: syntax error, unexpected tIDENTIFIER, expecting do or '{' or '('
x ? raise x : nil
```

Let me know if you'd like me to add more tests for those other cases.

### DIFF
--- a/librubyfmt/src/format.rs
+++ b/librubyfmt/src/format.rs
@@ -457,10 +457,7 @@ pub fn use_parens_for_method_call(
     }
 
     if name == "return" || name == "raise" || name == "yield" || name == "break" {
-        if ps.current_formatting_context() == FormattingContext::Binary {
-            return true;
-        }
-        if ps.current_formatting_context() == FormattingContext::IfOp {
+        if ps.current_formatting_context_requires_parens() {
             return true;
         }
         match args {
--- a/librubyfmt/src/format.rs
+++ b/librubyfmt/src/format.rs
@@ -460,6 +460,9 @@ pub fn use_parens_for_method_call(
         if ps.current_formatting_context() == FormattingContext::Binary {
             return true;
         }
+        if ps.current_formatting_context() == FormattingContext::IfOp {
+            return true;
+        }
         match args {
             ArgsAddStarOrExpressionList::ArgsAddStar(_) => return true,
             _ => return false,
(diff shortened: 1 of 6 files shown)


librubyfmt/src/format.rs
```diff
@@ -460,6 +460,9 @@ pub fn use_parens_for_method_call(
         if ps.current_formatting_context() == FormattingContext::Binary {
             return true;
         }
+        if ps.current_formatting_context() == FormattingContext::IfOp {
```
fables-tales: Thanks for this pull request! It's almost exactly what I would have done.

Would you mind collapsing this conditional and the one above to a method on parser state: `ps.current_formatting_context_requires_parens()` and then moving `Binary && IfOp` on to that object?

fables-tales (changes requested):

librubyfmt/src/format.rs
kaoru: Sounds good 😄 How's this? 90303fc6cb058ebedcc6be8c2d8a1d8e0e28447e

I'm writing Rust with absolutely no experience so it was a bit trial and error 😅 The tests and linter both seem happy though.

fables-tales (approved):

Issue #126, opened 2019-04-07
editor plugin for sublime


fables-tales:

toreriklinnerud:
Hi @samphippen I've just written a plugin for sublime: https://github.com/toreriklinnerud/sublime-rubyfmt 

If you agree I can prepare a PR so that it can be merged into this repository. It appears that this would prevent https://packagecontrol.io from assembling the extension directly from git. Instead we might need a little build script to assemble the .sublime-package and point them to that.

toreriklinnerud:
I've updated the plugin to reflect that rubyfmt is expected to be located in `~bin/rubyfmt.rb`

joe-sharp:
Initial Test works!:
```rb
def foo (foo,    bar)
   p      foo
  p  bar
  end
```
to =>
```rb
def foo(foo, bar)
  p(foo)
  p(bar)
end
```

joe-sharp:
I tested:

- Overriding the rubyfmt executable location in settings to point to my user's bin directory.
- Turning off auto-format on save
- Formatting via:
  - Auto-format on save
  - Format via Shortcut [Cmd + ;]
  - Format via Command Pallette

I am running:
macOS Mojave 10.14.6
Sublime Text 3.2.2 Build 3211

All tests passed so far. I am open to further testing suggestions. 😄

joe-sharp:
Filed https://github.com/toreriklinnerud/sublime-rubyfmt/issues/1 for some documentation issues.

fables-tales:
@joe-sharp could you provide a pull request that contains instructions for how to set up sublime to the readme?

PR #244, opened 2020-09-11
Add editor support entry for sublime text

toreriklinnerud:
As requested in https://github.com/penelopezone/rubyfmt/issues/126

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,28 @@ Once installed, add the following to vscode's `settings.json` file:
   },
 ```
 
-## RubyMine (and similar Jetbrains family IDE)
+### RubyMine (and similar Jetbrains family IDE)
 
 [Install](https://www.jetbrains.com/help/ruby/settings-tools-file-watchers.html) the File Watchers plugin and configue it like shown below.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,21 @@ Once installed, add the following to vscode's `settings.json` file:
   },
 ```
 
+### Sublime Text
+
+Install the [rubyfmt plugin](https://github.com/toreriklinnerud/sublime-rubyfmt/) from [Package Control](https://packagecontrol.io): Install Package -> rubyfmt. 
+
+`ruby` and `rubyfmt` are assumed to be on path. Ruby files are formatted on save or by pressing `Alt + ;` or on macOS: `Cmd + ;`
```
fables-tales: why does ruby need to be on the path?

fables-tales:
could you rebase? Sorry we're getting lots of readme updates right now

README.md
toreriklinnerud: I had the problem where rubyfmt would crash when there were syntax errors in the ruby file so if rubyfmt exits with an error I'm invoking `ruby -c` to check for syntax errors and surface them in the editor window to give the user an idea of why the file isn't being formatted.

toreriklinnerud:
Rebased

PR #256, opened 2020-10-23
Support class/module top_const_ref

clint-stripe:
This kinda threw the name of `ConstPathRefOrConstRef` out the window, which is a little weird. But a TopConstRef is kind of just a ConstPathRef if you squint? 😅


fables-tales:
@clint-stripe so I would rename this to `ConstPathRefOrConstRefOrTopConstRef`, would you mind doing that?

PR #264, opened 2020-12-09
Use colordiff in test runner when available

dbalatero:
This updates the test runner to prefer `colordiff` if it's in the `$PATH`. Falls back to the original `diff` behavior.

image

### DIFF
--- a/script/functions.sh
+++ b/script/functions.sh
@@ -21,7 +21,7 @@ diff_files() {
     ACTUAL=$1
     EXPECTED=$2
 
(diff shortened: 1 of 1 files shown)


dbalatero:
Alternatively, I can just `alias diff=colordiff`

fables-tales:
could make it reverse the way the + and - are, they're actually the wrong
way around in this (in my head)

dbalatero:
Would flipping the filename arguments to diff fix that?

fables-tales:
yes!

dbalatero:
Back to you @penelopezone !

PR #275, opened 2021-01-25
WIP: Handle heredoc strings in expressions

phiggins:
Hi,

I've made progress resolving some issues I found with rubyfmt's handling of heredocs, but I've reached a point where I can't find a solution for the problems I've found without changing some existing behavior, so this seems like a good point to ask for guidance in how to proceed. To avoid burying the lede, the problem is that there is not enough information present at the time of walking the Ripper tree to know what to do with the content of heredoc strings.

I initially found the problem trying to lint this file: https://github.com/chef/chef/blob/1899968131a27f703deac7de8f186abcf12da86b/lib/chef/resource/chef_client_config.rb#L131-L139

If I extract the troublesome portion to a file:
```
property :ssl_verify_mode, [Symbol, String],
  equal_to: %i{verify_none verify_peer},
  coerce: proc { |x| string_to_symbol(x) },
  description: <<~DESC
  Set the verify mode for HTTPS requests.

  * Use :verify_none for no validation of SSL certificates.
  * Use :verify_peer for validation of all SSL certificates, including the #{ChefUtils::Dist::Server::PRODUCT} connections, S3 connections, and any HTTPS remote_file resource URLs used in #{ChefUtils::Dist::Infra::PRODUCT} runs. This is the recommended setting.
  DESC
```

... this is the output when passed to rubyfmt:
```
property(
  :ssl_verify_mode,
  [Symbol, String],
  equal_to: %i[verify_none verify_peer],
  coerce: proc { |x| string_to_symbol(x) },
  description: <<~DESC
)
Set the verify mode for HTTPS requests.

* Use :verify_none for no validation of SSL certificates.
* Use :verify_peer for validation of all SSL certificates, including the #{ChefUtils::Dist::Server::PRODUCT}connections, S3 connections, and any HTTPS remote_file resource URLs used in #{ChefUtils::Dist::Infra::PRODUCT}runs. This is the recommended setting.
DESC
```

I initially thought it was a problem with the delimiter being printed to early, but a more focused example shows the problem is that heredoc string content is output too late. For example, this code:

```
puts <<~STRING, <<-STRING2,
    yellow
    fruit
    peel
  STRING
    cheese
    noodle
    sauce
  STRING2
  :burrito,
  <<~STRING3,
    jasmine
    earl grey
    oolong
  STRING3
  :rice
```

... produces this output:
```
puts(
  <<~STRING,
  <<-STRING2,

  :burrito,
  <<~STRING3,

  :rice
)
jasmine
earl grey
oolong
    cheese
    noodle
    sauce
yellow
fruit
peel
STRING
STRING2
STRING3
```

A related problem can be demonstrated by modifying one of the fixtures(https://github.com/penelopezone/rubyfmt/blob/367d43bee80c4c9a720d1c30aad3a57e55e0ffab/fixtures/small/pathalogical_heredocs_actual.rb#L1-L6). This code:

```
a = <<EOD
part 1 of heredoc #{ "not a heredoc" + <<EOM } after brace before newline
eom part
EOM
part 2 of heredoc
EOD
```

... produces this output:
```
a = <<EOD
part 1 of heredoc #{"not a heredoc" + <<EOM} after brace before newline
part 2 of heredoc

eom part
EOM
EOD
```

As I stated above, the problem as I see it is that at the time of walking the Ripper tree there is not enough information to decide what to do with the content of heredoc strings. In this PR, I've tried to move that decision making to the point when the abstract tree built from the Ripper tree is converted into the concrete tree of tokens that get output. I'll leave comments inline where necessary to explain particular decisions made.

Some form of the code in this PR was able to correctly format the examples I mentioned above. I say some form because I've tried so many small changes trying to get things to work that it's hard to remember what the current state is sometimes. :sweat_smile: 

The remaining problems I have been trying to deal with:
* Heredoc strings suffer from the same problem solved for other multiline strings in #273. Attempts to fix this have led to creating other formatting problems, mostly around the `is_multiline` function on `BreakableEntry`, which is finicky with respect to the record of lines processed from the original source file. I believe some of the existing fixtures include extra newlines as a result of this bug, for example: https://github.com/penelopezone/rubyfmt/blob/367d43bee80c4c9a720d1c30aad3a57e55e0ffab/fixtures/small/pathalogical_heredocs_expected.rb#L12-L18
* The attempt to solve #273 for heredocs by checking for the presence of newlines in an expression led to problems with maintaining whitespace from the original source file, for example not preserving multiline arrays in this fixture: https://github.com/penelopezone/rubyfmt/blob/367d43bee80c4c9a720d1c30aad3a57e55e0ffab/fixtures/small/return_actual.rb

I'm experienced with ruby but currently learning rust so I apologize for the rough code. :sweat_smile:  I realize this description is a bit scattered but I'm trying to coherently capture a bunch of random thoughts I've had while noodling on this for a while. Hopefully this makes some sense. This PR includes the commits from #273 and #274 to build off of the fixes from those PR.

Any advice on how to proceed would be greatly appreciated. Even if this is not something you want to take on at this time I've had fun and learned a lot about rust poking around and trying things. :heart:

### DIFF
--- a/fixtures/small/heredoc_param_actual.rb
+++ b/fixtures/small/heredoc_param_actual.rb
@@ -1,0 +1,22 @@
+puts <<~STRING, <<-STRING2,
+    yellow
+    fruit
+    peel
+  STRING
+    cheese
+    noodle
+    sauce
+  STRING2
+  :burrito,
+  <<~STRING3,
+    jasmine
+    earl grey
+    oolong
+  STRING3
+  :rice
+
+puts [:one, <<~TWO, :three], :four, [:five, <<~SIX, :seven], :eight
+two
+TWO
+six
+SIX
--- a/librubyfmt/src/line_tokens.rs
+++ b/librubyfmt/src/line_tokens.rs
@@ -1,4 +1,4 @@
-use crate::render_targets::{AbstractTokenTarget, BreakableEntry, ConvertType};
+use crate::render_targets::{AbstractTokenTarget, BreakableEntry, ConvertType, HeredocString};
 
 // represents something that will actually end up as a ruby token, as opposed to
 // something that has to be transformd to become a ruby token
@@ -120,6 +120,13 @@ impl ConcreteLineToken {
     pub fn is_in_need_of_a_trailing_blankline(&self) -> bool {
         self.is_conditional_spaced_token() && !self.is_block_closing_token()
     }
+
+    pub fn contains_newline(&self) -> bool {
+        match self {
+            Self::LTStringContent { content } => content.contains('\n'),
+            _ => false,
+        }
+    }
 }
 
 impl From<ConcreteLineToken> for ConcreteLineTokenAndTargets {
@@ -137,6 +144,9 @@ impl Into<AbstractLineToken> for ConcreteLineTokenAndTargets {
             ConcreteLineTokenAndTargets::ConcreteLineToken(clt) => {
                 AbstractLineToken::ConcreteLineToken(clt)
             }
+            ConcreteLineTokenAndTargets::HeredocString(hs) => {
+                AbstractLineToken::HeredocString(hs)
+            }
         }
     }
 }
@@ -145,6 +155,7 @@ impl Into<AbstractLineToken> for ConcreteLineTokenAndTargets {
 pub enum ConcreteLineTokenAndTargets {
     ConcreteLineToken(ConcreteLineToken),
     BreakableEntry(BreakableEntry),
+    HeredocString(HeredocString)
 }
 
 impl ConcreteLineTokenAndTargets {
@@ -170,6 +181,7 @@ impl ConcreteLineTokenAndTargets {
                 .fold("".to_string(), |accum, tok| {
                     format!("{}{}", accum, tok.into_ruby())
                 }),
+            Self::HeredocString(_hs) => "".to_string(),
             Self::ConcreteLineToken(clt) => clt.into_ruby(),
         }
     }
@@ -183,6 +195,7 @@ pub enum AbstractLineToken {
     SoftNewline,
     SoftIndent { depth: u32 },
     BreakableEntry(BreakableEntry),
+    HeredocString(HeredocString),
 }
 
 impl AbstractLineToken {
@@ -203,6 +216,7 @@ impl AbstractLineToken {
             }
             Self::ConcreteLineToken(clt) => ConcreteLineTokenAndTargets::ConcreteLineToken(clt),
             Self::BreakableEntry(be) => ConcreteLineTokenAndTargets::BreakableEntry(be),
+            Self::HeredocString(hs) => ConcreteLineTokenAndTargets::HeredocString(hs),
         }
     }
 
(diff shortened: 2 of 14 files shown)


fixtures/small/heredoc_param_actual.rb
```diff
@@ -0,0 +1,22 @@
...
+    earl grey
+    oolong
+  STRING3
+  :rice
+
+puts [:one, <<~TWO, :three], :four, [:five, <<~SIX, :seven], :eight
+two
+TWO
+six
+SIX
```
phiggins, line 22: :thought_balloon: This part of the test did not expose any additional problems with trunk-rubyfmt, but it did expose a problem with an earlier iteration of this PR so I left it in.

librubyfmt/src/render_queue_writer.rs
```diff
@@ -18,21 +18,59 @@ impl RenderQueueWriter {
...
+                        Self::format_heredocs(accum, heredocs);
+
+                        if !parts[1].is_empty() {
+                            accum.push(ConcreteLineToken::LTStringContent {
+                                content: parts[1].to_string()
+                            });
+                        }
+                    }
+
+                }
```
phiggins, line 62: :thought_balloon: This match branch was a late attempt to solve some of the issues mentioned. In a previous iteration of this code, the splitting of string tokens into strings and newlines was handled in `emit_string_content` in ParserState, which I think was a more elegant way to do this.

librubyfmt/src/render_queue_writer.rs
```diff
@@ -18,21 +18,59 @@ impl RenderQueueWriter {
...
-                    Self::format_breakable_entry(accum, be)
+                    Self::format_breakable_entry(accum, be, heredocs)
+                }
+                ConcreteLineTokenAndTargets::HeredocString(hs) => {
+                    heredocs.push(hs);
+                    #[cfg(debug_assertions)]
+                    {
+                        debug!("render_as: got heredoc {}", heredocs.len())
+                    }
+                }
```
phiggins, line 42: The match branches for `HeredocString` here and `ConcreteLineToken` below are doing the majority of the work in this PR. At this point in the app, there is enough information available to know when to emit the heredocs, ie, it has been decided if the expression will be single or multiline.

librubyfmt/src/line_tokens.rs
```diff
@@ -170,6 +181,7 @@ impl ConcreteLineTokenAndTargets {
                 .fold("".to_string(), |accum, tok| {
                     format!("{}{}", accum, tok.into_ruby())
                 }),
+            Self::HeredocString(_hs) => "".to_string(),
```
phiggins, line 184: :thought_balloon: This is technically correct but kind of a kludge. This would need to be revisited after more stuff was finalized.

fables-tales:
hi! I promise to actually look at this soon :). Open source burn-out is real but I will get to this when I can

phiggins:
No worries, I had assumed you were taking a break. :)

elliotlarson:
@penelopezone I just wanted to say, thank you for your work on this.

phiggins:
@penelopezone I saw your tweet about looking for maintainers for rubyfmt and I'd definitely be interested in helping. I didn't get a reply on twitter so I wanted to make sure this didn't fall through the cracks. :)

phiggins:
I might revisit this to see if there's anything else worth salvaging, but this has been superseded by #282.

Issue #276, opened 2021-04-10
Reenable macos build in CI


phiggins:
I disabled the macos build in CI in #269. It is failing with this error message:

```
   configure: error: cannot find required auxiliary files: config.guess config.sub
```

The full logs are here: https://github.com/penelopezone/rubyfmt/runs/2310759634

This is the same error that Ruby's own CI started throwing when it upgraded to autotools 2.70+: https://redmine.ruby-lang.org/issues/17723

Fixing this will likely involve some combination of waiting for that fix for ruby's build process to be backported to a 2.x branch, updating the version of ruby embedded in this project, and/or tweaking the github actions config to workaround the broken bits.

fables-tales:
can you try the `autoreconf -i` trick mentioned in that thread?

phiggins:
I did try that, I got a different error message. I think this is it but I can start a new PR to repro: https://github.com/penelopezone/rubyfmt/runs/2311550332?check_suite_focus=true#step:11:1086

phiggins:
The current status of this:

* I opened #277 to see if changing `autoconf` to `autoreconf -i` would fix things as indicated in that ruby issue, and it didn't. It's currently failing with this error message:

```
  /Users/runner/work/rubyfmt/rubyfmt/librubyfmt/ruby_checkout/ruby-2.6.6//configure: line 4103: syntax error near unexpected token `fi'
  /Users/runner/work/rubyfmt/rubyfmt/librubyfmt/ruby_checkout/ruby-2.6.6//configure: line 4103: `fi ;; #('
```
* I'll keep that branch open to investigate potential solutions. If I can figure out a non-grossly hacky solution that makes it go green I'll merge it.
* I poked around to see if anyone else was having this problem and didn't, but I also couldn't find any references to anyone building Ruby 2.6.x with autoconf 2.71 either. For instance, ruby-build did the `autoreconf -i` change but it's not clear to me that it rebuilds all the rubies in their CI: https://github.com/rbenv/ruby-build/pull/1743

Things to investigate:

* Homebrew does not usually support installing multiple versions, but in this case it does have an older version of autoconf, 2.31, available (https://formulae.brew.sh/formula/autoconf). Pinning to an old version is not a great longterm solution but is better than the current state of a disabled mac build. I'll need to try installing the older version and getting the build system to use it, but I'm only familiar with the basics of homebrew and am not sure how easy it will be to debug this via CI.
[...]

PR #277, opened 2021-04-10
Reenable macos build in CI

phiggins:
#276


fables-tales:
seems better than not, maybe we can provide build instructions for OS X to folks, but this is weird and sad

phiggins:
@penelopezone I would expect a backport of the fix for this, because at it is no one with the new autoconf can build any released ruby which seems like a big problem for mac users or anyone on a rolling release linux distro.

PR #282, opened 2021-04-18
Fixes rendering heredocs in method calls.

### LINKED ISSUE #209: Error when formatting a heredoc passed as a hash value
* Ruby version: 2.6.3
* Rubyfmt git sha: 9d65851d1ba5ab44889624cd7122a28d5f1e81e2

## Input file

```ruby
hash = {
  msg: <<~EOD,
    heredoc msg
  EOD
  opts: {
    foo: 'bar'
  }
}
```

## Rubyfmt's output

``` ruby
hash = {
  msg: <<~EOD,
  opts: {
    foo: "bar"
  }
}
heredoc msg
EOD
```

## This breaks because

It pulls the remaining keys of the hash into the heredoc instead.

fables-tales:
The problem was that essentially they would always be placed in the
wrong place due to how we assumed they work with newlines. This puts the
heredocs on collapsing newlines if we go to render an abstract target
(like a method call) and then pops them in the right place at the render
queue writer stage.

Fixes #209

### DIFF
--- a/ci/string_literals_stress_test.rb
+++ b/ci/string_literals_stress_test.rb
@@ -36,3 +36,47 @@ puts %q(\\"\\")
 puts %q(\))
 puts %Q(\))
 puts %<foo\>>
+
+puts(
+  1,
+  2,
+  <<~TXT,
+    3
+    4
+  TXT
+)
+
+puts(
+  1,
+  2,
+  <<TXT,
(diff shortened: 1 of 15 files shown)


phiggins:
Nice! I'll check if the examples I added in #275 are still relevant.

Issue #221, opened 2020-07-10
"Rubyfmt failed to correctly deserialize a tree from ripper" in concurrent-ruby gem


ivoanjo:
Hello again! After [my previous issue report](https://github.com/penelopezone/rubyfmt/issues/220) I decided to try `rubyfmt` with a few other Ruby gems that I had checked out locally.

`rubyfmt` fails to process multiple files from `concurrent-ruby`.

* Ruby version: `ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]`
* Rubyfmt git sha: https://github.com/penelopezone/rubyfmt/releases/tag/v0.3.1

## Input file

All of these fail:

* file was: docs-source/actor/messaging.in.rb
* file was: docs-source/actor/messaging.out.rb
* file was: examples/a-tour-of-go-channels/buffered-channels.rb
* file was: examples/a-tour-of-go-channels/channels.rb
* file was: examples/a-tour-of-go-channels/equivalent-binary-trees.rb
* file was: examples/a-tour-of-go-channels/select.rb
* file was: examples/go-by-example-channels/channel-directions.rb
* file was: examples/go-by-example-channels/channel-synchronization.rb
* file was: examples/go-by-example-channels/closing-channels.rb
* file was: examples/go-by-example-channels/rate-limiting.rb
* file was: examples/go-by-example-channels/timers.rb
* file was: examples/go-by-example-channels/worker-pools.rb
* file was: lib/concurrent-ruby/concurrent/collection/map/atomic_reference_map_backend.rb
* file was: lib/concurrent-ruby/concurrent/thread_safe/util/striped64.rb
* file was: spec/concurrent/atomic/count_down_latch_spec.rb
* file was: spec/concurrent/atomic/cyclic_barrier_spec.rb
* file was: spec/concurrent/atomic/event_spec.rb
* file was: spec/concurrent/mvar_spec.rb
* file was: tmp/x86_64-linux/stage/lib/concurrent/atomic_reference/direct_update.rb

Link to concurrent-ruby repository: https://github.com/ruby-concurrency/concurrent-ruby

## Rubyfmt's output

(None)

## This breaks because

Does not work ;)

phiggins:
With #280, #281, #283, and #285 merged I'm able to successfully run rubyfmt on concurrent-ruby. Thanks for the report!

fables-tales:
@phiggins this is amazing TY so much <3

[...]

PR #283, opened 2021-04-18
Handle flags in calls to rspec DSL methods.

phiggins:
Another one from looking at #221, this caused rubyfmt to panic with this error message:

```
thread 'main' panicked at 'failed to convert', librubyfmt/src/parser_state.rs:754:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/librubyfmt/src/format.rs
+++ b/librubyfmt/src/format.rs
@@ -678,7 +678,12 @@ pub fn format_list_like_thing_items(
             //raise "this is bad" if expr[0] == :tstring_content
 
             if single_line {
-                format_expression(ps, expr);
+                match expr {
+                    Expression::BareAssocHash(bah) => {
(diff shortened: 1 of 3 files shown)


librubyfmt/src/format.rs
```diff
@@ -678,7 +678,12 @@ pub fn format_list_like_thing_items(
             //raise "this is bad" if expr[0] == :tstring_content
```
fables-tales, line 678: we should probably delete this line lol

phiggins: I can delete that whole comment block if you don't think its relevant context anymore.

phiggins: Done in #284

phiggins:
... and this only fixes the case where there is exactly one item in the hash lol. I might be out of open sourcing time today but I'll revisit this soon.

PR #287, opened 2021-05-09
Variable tracking

fables-tales:
PTAL @phiggins

### DIFF
--- a/fixtures/small/variable_binding_actual.rb
+++ b/fixtures/small/variable_binding_actual.rb
@@ -14,4 +14,9 @@ def bees
   p head
 end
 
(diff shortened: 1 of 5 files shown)


phiggins (approved):
LGTM! :+1: 

I thought that you might be able to fool this by mixing `self.` in with bare calls, but it handled everything I could come up with.

fables-tales:
https://github.com/penelopezone/rubyfmt/pull/287/files#diff-63b92541cacb647f9227c6eac250e906cb60dade9a39a892517f2d99ec122a28R581-R585 that's what this handles!

Issue #233, opened 2020-08-29
Create a cargo workspace containing librubyfmt


sgrif:
I keep running `cargo fmt` and `cargo clippy` in the main directory forgetting most of the code is in a subdirectory. Having the outer directory be a workspace would fix this

fables-tales:
Sure, could you send a patch and update `make fmt` to do this?

phiggins:
Closed by #290.

PR #290, opened 2021-05-18
Use a cargo workspace.

phiggins:
Implements #233.

This doesn't seem like a huge change but makes building the project and using all the `cargo` commands a little simpler. 

* [The workspaces section of the cargo book was a useful reference](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)
* [The workspaces section of the rust book was less useful](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspace-section), it made it sound like only one option, what the cargo book calls the `virtual manifest`, was possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,34 +12,26 @@ endif
 LDFLAGS +=  -lz
 
 all: submodules release debug
-debug: target/debug/librubyfmt.a target/debug/rubyfmt-main
-release: target/release/librubyfmt.a target/release/rubyfmt-main
 
-target/debug/rubyfmt-main: librubyfmt/src/*.rs librubyfmt/Cargo.toml src/*.rs Cargo.toml
+debug:
 	cargo build
 
-target/release/rubyfmt-main: librubyfmt/src/*.rs librubyfmt/Cargo.toml src/*.rs Cargo.toml
+release:
 	cargo build --release
 
 submodules:
 	git submodule init
 	git submodule update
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,34 +12,26 @@ endif
...
-	mkdir -p target/release
-	cd librubyfmt && cargo build --release
-	cp librubyfmt/target/release/librubyfmt.a $@
+target/release/deps/librubyfmt-*.a: release
 
-target/debug/librubyfmt.a: librubyfmt/src/*.rs librubyfmt/Cargo.toml
-	mkdir -p target/debug
-	cd librubyfmt && cargo build
-	cp librubyfmt/target/debug/librubyfmt.a $@
+target/debug/deps/librubyfmt-*.a: debug
```
phiggins, line 34: I was concerned that using a wildcard for these would be a problem, but I did a few cycles of making a small change and rebuilding and there was only ever one file matching this pattern at once.

phiggins:
This is not building on CI for some reason and I'm out of time to investigate for now.

phiggins:
CI seems busted for now, I'll keep #291 open to investigate / fix.

phiggins:
This is passing on CI now so I've marked it as ready for review.

@penelopezone I think everything is still working how it was and this seems like an improvement. WDYT?

fables-tales (approved):

PR #292, opened 2021-05-29
Formatting lock testing

fables-tales:
@phiggins PTAL

This implements what I'm going to call "formatting lock" testing. Basically I think Rubyfmt is growing to the point now where we format ~most ruby files kind of correctly. As we progress on weirder edge cases, it'd be nice to know we're not backsliding, so this "locks" formatting of particular files in place.

I'd love input on the format, the scripts, etc as a first pass code review, but *also*, I'd love if you could contribute repos/files to this. I think the goal is if we can get to 100% coverage on a few popular repos, we can cut a first pass "daily driver" build of Rubyfmt.


script/tests/test_formatting_locks.sh
```diff
@@ -0,0 +1,31 @@
...
+        then
+            echo "$fn did not match"
+            exit 1
+        fi
+
+    done
+    )
+done
+echo 'everything checks out'
+
```
phiggins, line 31: The idea with this is that we recognize files in the wild that demonstrate some particular behavior of the formatter and add them to the test suite? I like it!

I'm curious though why this wouldn't reuse the mechanism already used for fixtures, when this seems like a very similar method of testing. This script would provide a lot less information in the case of a failure, ie, it would say there was a failure but not what the diff was, and it would be more work to go about finding out what the change was. The time taken to run these tests would also grow linearly with the number of repos added, since there would need to be a git checkout for each versus including the individual files in this repo.

script/tests/test_formatting_locks.sh
```diff
@@ -0,0 +1,31 @@
...
+set -euo pipefail
+source ./script/functions.sh
+REPO_DIRS=$(find ./formatting_locks -type d -mindepth 2)
+for REPO_DIR in $REPO_DIRS
+do
+    HEAD=$(cat "$REPO_DIR/HEAD")
+    FILES=$(cat "$REPO_DIR/files")
+    REPO=$(echo "$REPO_DIR" | sed 's/.\/formatting_locks\///')
+    (
+    cd "$(mktemp -d)"
```
phiggins, line 11: Could this use a static directory instead of a new temp dir each time? If this is checking out the same versions each time it seems wasteful to repeat that for each run of this script when that is easily cacheable.

script/tests/test_formatting_locks.sh
phiggins: I've been meaning to ask how you'd feel about rewriting some of the test scripts in (probably) ruby and this seems like a decent chance to do so. I've written my fair share of bash scripts and when they are around the size of this one and the fixture testing script is when I get the feeling I should rewrite them in something like ruby.

Not entirely related, but I've been frustrated with the "fail first" style of running tests being used for the fixtures, because I don't often just want to know if anything failed but what all of the failures are. Running the tests with a full fledged test runner would make it easier to turn that bit on or off when appropriate.

phiggins:
I got my second vaccine dose the other day and am still feeling out of it but wanted to reply in a timely fashion. Hopefully these comments make sense! I'd be happy to discuss anything I brought up.

script/tests/test_formatting_locks.sh
fables-tales: > I've been meaning to ask how you'd feel about rewriting some of the test scripts in (probably) ruby and this seems like a decent chance to do so. I've written my fair share of bash scripts and when they are around the size of this one and the fixture testing script is when I get the feeling I should rewrite them in something like ruby.
> 
> Not entirely related, but I've been frustrated with the "fail first" style of running tests being used for the fixtures, because I don't often just want to know if anything failed but what all of the failures are. Running the tests with a full fledged test runner would make it easier to turn that bit on or off when appropriate.

As a former Ruby test framework maintainer, I wholly endorse this. (please use RSpec)

fables-tales: > I'm curious though why this wouldn't reuse the mechanism already used for fixtures, when this seems like a very similar method of testing. This script would provide a lot less information in the case of a failure, ie, it would say there was a failure but not what the diff was, and it would be more work to go about finding out what the change was. The time taken to run these tests would also grow linearly with the number of repos added, since there would need to be a git checkout for each versus including the individual files in this repo.

The idea here is that fixtures should be samples we've manually created and checked in to exhibit specific test cases in most places. Formatting locks are moreso designed to be a quick check against O(lots) of ruby, and so we use this mechanism so we don't have to check all that code in

script/tests/test_formatting_locks.sh
fables-tales: > Could this use a static directory instead of a new temp dir each time? If this is checking out the same versions each time it seems wasteful to repeat that for each run of this script when that is easily cacheable.

Yeah I think that's reasonable

script/tests/test_formatting_locks.sh
phiggins: > As a former Ruby test framework maintainer, I wholly endorse this. (please use RSpec)

I was going to use minitest/spec just so there wasn't an external dependency / `bundle install` step, haha. I'll play around with using rspec and see what I come up with.

phiggins: > Formatting locks are moreso designed to be a quick check against O(lots) of ruby, and so we use this mechanism so we don't have to check all that code in

Ok, I might be missing something because of the small numbers of samples so far, so I'll trust your judgement. :+1:

PR #299, opened 2021-06-01
Handle intentionally bad code

phiggins:
For whatever reason the chef repo has files with an .rb extension with garbage text: https://github.com/chef/chef/blob/master/spec/data/cookbooks/ignorken/recipes/ignoreme.rb

This was getting into the `on_string_literal` method with only one element on the string_delimiter stack, making `start_delim` nil on line 217: https://github.com/penelopezone/rubyfmt/blob/641534ce3730eaf060115a0c59b6bbbdd55ef4c0/librubyfmt/rubyfmt_lib.rb#L217

Before:
```
$ cat ../chef/spec/data/cookbooks/ignorken/recipes/ignoreme.rb | ./target/release/rubyfmt-main
A ruby error occurred: #<NoMethodError: undefined method `start_with?' for nil:NilClass>, please file a bug report at https://github.com/penelopezone/rubyfmt/issues/new
```

After:
```
$ cat ../chef/spec/data/cookbooks/ignorken/recipes/ignoreme.rb | ./target/release/rubyfmt-main
stdin contained invalid ruby syntax
```

### DIFF
--- a/librubyfmt/rubyfmt_lib.rb
+++ b/librubyfmt/rubyfmt_lib.rb
@@ -213,7 +213,7 @@ class Parser < Ripper::SexpBuilderPP
       end_delim = @string_stack.pop
       start_delim = @string_stack.pop
 
-      if start_delim != "\""
+      if start_delim && end_delim && start_delim != "\""
         if start_delim == "'" || start_delim.start_with?("%q")
           # re-evaluate the string with its own quotes to handle escaping.
           if args[0][1]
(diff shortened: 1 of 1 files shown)


sgrif (approved):
It seems odd that ripper puts us in `on_string_literal` in this case, but this seems like a reasonable fix

phiggins:
This does fix the crash, but I wanted to see if I could add a regression test for this and maybe refactor the code a little.

phiggins:
It's interpreting `%$` as the opening delimiter for a string delimited with two `$` but since there's only one in the file the delimiter stack only has one element. I tried a few things like `'` and it failed in the same way. This feels like it might be a bug in ripper (why trigger events when you know it's invalid code?) but its easy enough to workaround.

PR #303, opened 2021-06-01
Automatically inform the user if a new version is available

sgrif:
(This PR includes #302, as that refactor was done to make this PR easier to write)

Penelope asked me to "build an auto-update checking mechanism that
doesn't slow it down". This should meet that criteria, though not
slowing things down definitely made it trickier.

Once this PR is merged, the version field in `Cargo.toml` will need to
start getting updated whenever a new release is put out. I've also
updated the version field for librubyfmt, but it's not used by the code
in this commit.

To determine what the most recent version is, we hit GitHub's API to get
the tags on the repo, and look for the highest tag that looks like a
version (using semver sorting, not lexographic sorting which is what
GitHub would give us). To avoid spamming their API, we perform this
check at most once per day.

The difficult thing about checking for updates this way is that rubyfmt
is very likely to finish doing its thing faster than we can make that
network request, especially if the user is on a high latency connection.

Because of this, we spawn a detached child process to make the API
request. The version it finds is written to a file, and the mtime of
that file is used to track when the last time we asked GitHub was. The
only work done by the main rubyfmt process is reading that file.

As written, this commit does technically slow it down since I'm blocking
on the file I/O. I believe the impact here should be negligible, though
it might be measurable if rubyfmt is reading from stdin/out instead of
doing file I/O elsewhere. If this is a concern, I can move the file IO
to another thread and have it race, so if we can't read the file to
check the version before formatting finishes, we just exit. I think the
gains from doing this would be marginal at best.

Since we need to have persistent state across runs, the question arises
of where we should actually put this file. I opted to put this in a data
dir, rather than a cache dir, since we don't actually want this file
getting cleaned up if disk space is low. On Linux this will end up in
either `$XDG_DATA_HOME` or `~/.local/store`. On Mac it'll end up in
`~/Library/Application Support`. On Windows it'll end up in
`~/AppData/Roaming`.

Because we're hitting an external API for something that is strictly an
optional feature, I've been very defensive about making sure we hit the
API *at most* once per day. We pessimistically update the mtime before
actually making the request, so that if something goes wrong we don't
end up spamming their API after every run. If an error does occur for
some reason, it will be silently ignored in release mode. In debug
builds it will panic if the error happened in the parent process, or
print to stderr if it happened in the child.

This uses `ureq`, which is definitely not a common or standard HTTP
client. I would normally go with `reqwest`, but that ends up pulling in
the entire tokio ecosystem, since it's async by default. I didn't want
to add that sort of impact on our compile times. `ureq` is a lighter
weight HTTP client that suits our needs, and is widely used enough that
I'm comfortable with depending on it.

### DIFF
--- a/librubyfmt/Cargo.toml
+++ b/librubyfmt/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rubyfmt"
-version = "0.1.0"
+version = "0.6.8-pre"
 authors = ["Penelope Phippen <penelopedotzone@gmail.com>"]
 edition = "2018"
 build = "build.rs"
@@ -10,7 +10,6 @@ build = "build.rs"
 [dependencies]
 serde = { version = "1.0", features=["derive"] }
 serde_json = "1.0.40"
-bytecount = "0.6.0"
 backtrace = "0.3.45"
 libc = "0.2.68"
 ripper_deserialize = { path = "ripper_deserialize" }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,7 @@
 #![deny(warnings, missing_copy_implementations)]
 
+mod updates;
+
 use std::ffi::{OsStr, OsString};
 use std::fs::{self, metadata, read_to_string, OpenOptions};
 use std::io::{self, Read, Write};
@@ -111,6 +113,7 @@ fn handle_error_from(err: rubyfmt::RichFormatError, source: &Path, error_exit: E
 }
 
 fn main() {
+    updates::begin_checking_for_updates();
     let res = rubyfmt::rubyfmt_init();
     if res != rubyfmt::InitStatus::OK as libc::c_int {
         panic!(
@@ -142,6 +145,9 @@ fn main() {
             eprintln!("{}", include_str!("../README.md"));
             exit(1);
         }
+        (Some("--internal-fetch-latest-version"), _) => {
+            updates::fetch_latest_version().unwrap();
+        }
         // Single file
         (_, [filename]) => {
             if let Ok(md) = metadata(&filename) {
@@ -168,4 +174,5 @@ fn main() {
             format_parts(parts);
         }
     }
+    updates::report_if_update_available();
 }
--- a/src/updates.rs
+++ b/src/updates.rs
@@ -1,0 +1,128 @@
+use std::env;
+use std::error::Error;
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+#[cfg(not(debug_assertions))]
+use std::process::Stdio;
+use std::time::Duration;
+
+const ONE_DAY: Duration = Duration::from_secs(60 * 60 * 24);
+
+#[derive(Debug, Clone, serde::Deserialize)]
+struct GithubTag {
+    name: String,
+}
+
+/// Begins the process of checking if a new version is available
+///
+/// This will check the timestamp of a file containing the latest version of rubyfmt. If the file
+/// hasn't been updated in at least a day, the file will be touched and a process will be spawned
+/// to write the most recent version.
+///
+/// Panics on any error in debug builds. Since this process is strictly optional, any errors are
+/// ignored in release builds.
+pub(crate) fn begin_checking_for_updates() {
+    let result = try_begin_checking_for_updates();
+    if cfg!(debug_assertions) {
+        result.unwrap();
+    }
+}
+
+fn try_begin_checking_for_updates() -> io::Result<()> {
+    let path = path_to_latest_version_file()?;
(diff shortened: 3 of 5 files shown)


src/updates.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// Determine the most recent version of Rubyfmt, and write it to a file
+///
+/// Because GitHub paginates its responses to this endpoint, this assumes that the latest version
+/// is in the first 30 tags given to us (which are sorted in lexographically descending order).
+/// Tags which are a version release are assumed to always be `v` followed by a semver version (for
+/// example, v1.0.0)
+pub(crate) fn fetch_latest_version() -> Result<(), Box<dyn Error>> {
+    let max_version = ureq::get("https://api.github.com/repos/penelopezone/rubyfmt/tags")
```
penelope-stripe, line 94: can this be releases instead of tags?

sgrif: I'll check. If we can do you want it to always be the latest release or should I continue parsing the name and grab the highest semver version as insurance against releasing a security patch for an old version

sgrif: So it looks like I *can* hit the releases endpoint instead of tags, but I'm just going to have to end up using the tag name anyway since the release name hasn't been consistent. I don't think there's any benefit to doing this since the releases endpoint is effectively just the tags endpoint with a much larger payload size in this case, but I'm happy to write the code if that's what you want

src/main.rs
```diff
@@ -110,60 +114,66 @@ fn handle_error_from(err: rubyfmt::RichFormatError, source: &str, error_exit: Er
...
-    } else if args[1] == "-i" {
-        // inline a file or directory
-        let parts = &args[2..args.len()];
-        format_parts(parts);
-    } else {
-        // inline many files and directories
-        let parts = &args[1..args.len()];
-        format_parts(parts);
+        // Multiple files
+        (Some("-i"), [_, parts @ ..]) | (_, parts) => {
```
fables-tales: What does the `@ ..` operator do?

librubyfmt/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rubyfmt"
-version = "0.1.0"
+version = "0.6.7"
```
fables-tales: should this `0.6.8.pre` maybe? the under development version is not version 0.6.7 (the pre string is how we did it in rspec)

src/main.rs
```diff
@@ -38,16 +41,17 @@ fn rubyfmt_file(file_path: PathBuf) -> Result<(), FileError> {
...
+            handle_error_from(e, &file_path, ErrorExit::NoExit);
             Ok(())
         }
     }
 }
 
-fn rubyfmt_dir(path: &str) {
-    for entry in glob(&format!("{}/**/*.rb", path)).expect("it exists") {
+fn rubyfmt_dir(path: &Path) {
+    // FIXME: Look for an implementation of glob that actually takes a proper base dir
```
fables-tales: do we have a corresponding issue for this?

src/updates.rs
```diff
@@ -0,0 +1,128 @@
...
+/// This will check the timestamp of a file containing the latest version of rubyfmt. If the file
+/// hasn't been updated in at least a day, the file will be touched and a process will be spawned
+/// to write the most recent version.
+///
+/// Panics on any error in debug builds. Since this process is strictly optional, any errors are
+/// ignored in release builds.
+pub(crate) fn begin_checking_for_updates() {
+    let result = try_begin_checking_for_updates();
+    if cfg!(debug_assertions) {
+        result.unwrap();
```
fables-tales, line 29: 👍

src/updates.rs
fables-tales: cool, I think this is fine as is 🚀

fables-tales (approved):
overall LGTM, had some questions but nothing here concerns me.

src/main.rs
sgrif: `ident @ pat` binds whatever matches `pat` to `ident`. `..` in a pattern is "the rest of it", which in this case is the rest of the slice.

librubyfmt/Cargo.toml
sgrif: That isn't a valid semver version. I could do `-pre`, but setting the version on git to something different like this is pretty uncommon in the Rust world, since if someone *does* want to point at git for some reason this makes it more difficult for them to do so

src/main.rs
sgrif: #308

librubyfmt/Cargo.toml
sgrif: Updated to `0.6.8-pre` for now, we can undo later if you want to stick with Rust norms

fables-tales (approved):

PR #306, opened 2021-06-03
Handle backtick symbol.

phiggins:
The Chef devs are playing a practical joke on me aren't they: https://github.com/chef/chef/blob/32686e39d4cda5071619997b44766937b67fa164/spec/unit/provider/package/zypper_spec.rb#L38


librubyfmt/src/format.rs
```diff
@@ -889,22 +889,37 @@ pub fn format_kw(ps: &mut dyn ConcreteParserState, kw: Kw) {
...
+    if ps.at_start_of_line() {
+        ps.emit_indent();
+    }
+
+    handle_string_and_linecol(ps, backtick.1, backtick.2);
+
+    if ps.at_start_of_line() {
+        ps.emit_newline();
+    }
+}
```
phiggins, line 902: This function has been copied at least 4 times handling different types and should get refactored.

PR #308, opened 2021-06-04
Replace `glob` with basic directory traversal

sgrif:
The `glob` crate as well as any alternatives I could find, don't provide
a way to provide a base directory. That means that in the unlikely event
that `rubyfmt` was called on a directory that is not valid UTF-8, we
error out. Ultimately the thing we were using `glob` for, was relatively
simple, and we don't need to pull in a dependency to save us ~3 lines of
code


phiggins (approved):
Looks good to me :+1:

PR #316, opened 2021-06-07
Remove ruby version from submodule path

sgrif:
As we update the version of Ruby over time we're not going to want to
update a bunch of hard coded version numbers. The `ruby_checkout/README`
file seemed redundant, as a git submodule is not the same as
redistributing code, so there shouldn't be any license or copyright
concerns. Because of that I've just moved the submodule up a level

Related to #315.

### DIFF
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -39,7 +39,7 @@ jobs:
     - uses: actions/cache@v2
       with:
         path: |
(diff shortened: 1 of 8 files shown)


sgrif:
Closing for now. Removing the hard coded path invalidates the cache on the checkout dir, which causes `autoconf` to need to be run, which fails on Mac. https://github.com/penelopezone/rubyfmt/issues/315#issuecomment-856297259

PR #318, opened 2021-06-10
Update embedded ruby from 2.6.6 -> 2.7.x

phiggins:
- Updates the version of ruby embedded in the repo to https://github.com/ruby/ruby/tree/29bbad939939c6dceb804aac667ba372fdee4ef5, the commit on the 2.7 branch that has the fixes needed for building with autoconf 2.7+.
- Fixes the cache key for the ruby checkout to (hopefully) invalidate the cache when the commit of the git submodule changes.
- Move the submodule to a version agnostic directory (`ruby_checkout/ruby-2.6.6` -> `ruby_checkout`).
- Remove macos autoconf hacks from CI config.

### DIFF
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -57,13 +57,4 @@ jobs:
         rustup component add clippy --toolchain stable-x86_64-apple-darwin
         rustup component add rustfmt --toolchain stable-x86_64-apple-darwin
         brew install shellcheck sccache
-
-        # Homebrew installs autoconf 2.71 which is incompatible with the
-        # version of ruby embedded in this project, so force the installation
-        # of the last known good version, 2.69.
-        brew unlink autoconf
-        brew tap-new $USER/local-autoconf
-        brew extract --version=2.69 autoconf $USER/local-autoconf
-        brew install autoconf@2.69
-        echo "/usr/local/opt/autoconf@2.69/bin" >> $GITHUB_PATH
     - run: ./script/ci
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -40,7 +40,7 @@ jobs:
       with:
         path: |
           librubyfmt/ruby_checkout/ruby-2.6.6
-        key: ${{ runner.os }}-ruby26-full
+        key: ${{ runner.os }}-ruby-${{ hashFiles('.git/modules/librubyfmt/ruby_checkout/ruby-2.6.6/HEAD') }}
     - if: runner.os == 'Linux'
       run: |
         sudo apt-get install -y shellcheck build-essential ruby-dev bison
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -56,5 +56,5 @@ jobs:
       run: |
         rustup component add clippy --toolchain stable-x86_64-apple-darwin
         rustup component add rustfmt --toolchain stable-x86_64-apple-darwin
-        brew install shellcheck sccache
+        brew install shellcheck sccache automake
     - run: ./script/ci
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -65,4 +65,5 @@ jobs:
         brew tap-new $USER/local-autoconf
         brew extract --version=2.69 autoconf $USER/local-autoconf
         brew install autoconf@2.69
+        echo "/usr/local/opt/autoconf@2.69/bin" >> $GITHUB_PATH
     - run: ./script/ci
--- a/.github/workflows/2.6.yml
+++ b/.github/workflows/2.6.yml
@@ -39,8 +39,8 @@ jobs:
     - uses: actions/cache@v2
       with:
         path: |
-          librubyfmt/ruby_checkout/ruby-2.6.6
-        key: ${{ runner.os }}-ruby-${{ hashFiles('.git/modules/librubyfmt/ruby_checkout/ruby-2.6.6/HEAD') }}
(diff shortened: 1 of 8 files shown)


librubyfmt/build.rs
```diff
@@ -10,7 +10,7 @@ fn main() -> Output {
     #[cfg(target_os = "linux")]
     let libname = "ruby-static";
     #[cfg(target_os = "macos")]
-    let libname = "ruby.2.6-static";
+    let libname = "ruby.2.7-static";
     #[cfg(all(target_arch = "x86_64", windows))]
     let libname = "x64-vcruntime140-ruby260-static";
```
sgrif: This needs to be updated as well

sgrif: I'm not very familiar with Makefiles, but if it provides some way to get the value of a variable within the makefile, it'd be great to use `RUBY_SO_NAME` here

.github/workflows/2.6.yml
```diff
@@ -64,4 +60,5 @@ jobs:
         brew tap-new $USER/local-autoconf
         brew extract --version=2.69 autoconf $USER/local-autoconf
         brew install autoconf@2.69
+        echo "/usr/local/opt/autoconf@2.69/bin" >> $GITHUB_PATH
```
sgrif: Since the autoconf fix has been merged, do we still need to do this? Can we just use autoconf 2.71 instead?

librubyfmt/build.rs
phiggins: I would love to figure out a better way to do this. Having to remember to change a bunch of strings all over the place sucks.

After we get this sorted out I would like to (eventually) work on getting a Windows CI worker going.

.github/workflows/2.6.yml
phiggins: We'd have to either track the 2.7 branch or cherry-pick that fix. This PR updates the git submodule to use the 2.7.3 tag specifically.

librubyfmt/build.rs
sgrif: Yes, I agree on both points. Though once we get rid of the version in the path, it's only 3-4 strings total so it's not the worst thing in the world. It absolutely sucks that the format changes all over the place though.

.github/workflows/2.6.yml
sgrif: What do you think about just pointing it at that commit instead of the commit pointed at by the 2.7.3 tag for now?

phiggins: That's a good idea. I can probably try that out in the next few days.

librubyfmt/build.rs
```diff
@@ -62,10 +62,10 @@ fn make_configure(ruby_checkout_path: &Path) -> Output {
             .status()?;
         check_process_success("make configure", o)
     } else {
-        let o = Command::new("autoconf")
```
sgrif, line 65: Does `autoreconf -i` behave the same as `make configure` if the inputs are unchanged? If so we can probably remove the `if` here

phiggins: Good call. I'm not sure what `make configure` does differently but it's not mentioned in ruby's docs on building so I'm inclined to delete it.

phiggins:
Using `autoreconf -i` in place of `autoconf` is mentioned here: https://redmine.ruby-lang.org/issues/17723

phiggins:
This is green again, pointing to the commit to the 2.7 branch with the autoconf fixes and using `autoreconf -i` as recommended by the ruby folks.

phiggins:
If we need to use the old autoconf again for whatever reason, homebrew has added a package for autoconf 2.69 so this wouldn't require the custom tap and all that.

phiggins:
TODOs:
- [x] Figure out a way to cache the ruby git submodule such that it's a cache miss when the commit it points to is changed.
- [x] Add @sgrif's path changes from #316.
- [ ] Smarter way to handle the shared lib name?
- [x] Update Windows shared lib names

librubyfmt/build.rs
sgrif: It's literally just

```
configure: configure.ac
    autoconf
```

fables-tales:
I appreciate that two of you for working on this ❤️

phiggins:
This is ready for review. @sgrif @penelopezone, anything left to be done here?

.github/workflows/ci.yml
```diff
@@ -17,6 +17,7 @@ jobs:
   CI_26:
```
fables-tales: should this change?

Makefile
```diff
@@ -2,7 +2,7 @@
 
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S), Darwin)
-	LDFLAGS=-framework Foundation
+	LDFLAGS=-framework Foundation -framework Security
```
fables-tales, line 5: just to check, does this need to be documented in the release readme?

.github/workflows/ci.yml
phiggins: Good catch!

Makefile
phiggins: Does it? I'm not sure what this does tbh. I got an error on CI and that was the suggested solution I found on the web.

sgrif: It shouldn't, these are mac specific and exist on all macs

sgrif: But also I'd like to see if we can yeet this Makefile at some point because it doesn't seem to do much that isn't already handled by Cargo

phiggins:
If there's no other comments I'll probably merge this tonight 🚀

sgrif (approved):

PR #323, opened 2021-06-14
Sync submodule in build.rs

phiggins:
Borrowed from here: https://github.com/rust-lang/git2-rs/blob/b055a28f0eebe491f4e2047be6dd3190e3c31060/libgit2-sys/build.rs#L23-L27

### DIFF
--- a/librubyfmt/build.rs
+++ b/librubyfmt/build.rs
@@ -20,6 +20,13 @@ fn main() -> Output {
 
     let path = std::env::current_dir()?;
     let ruby_checkout_path = path.join("ruby_checkout");
+
+    if !ruby_checkout_path.join(".git").exists() {
+        let _ = Command::new("git")
+            .args(&["submodule", "update", "--init"])
+            .status();
+    }
+
     make_configure(&ruby_checkout_path)?;
     run_configure(&ruby_checkout_path)?;
     build_ruby(&ruby_checkout_path)?;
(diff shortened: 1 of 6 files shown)


librubyfmt/build.rs
```diff
@@ -20,6 +20,13 @@ fn main() -> Output {
 
     let path = std::env::current_dir()?;
     let ruby_checkout_path = path.join("ruby_checkout");
+
+    if !ruby_checkout_path.join(".git").exists() {
```
sgrif: Do we actually need this conditional? Doesn't git already handle this?

sgrif (approved):

librubyfmt/build.rs
phiggins: I copied and pasted this but probably not. In the worst case it runs the command when it doesn't need to and your build fails because github is down, but that seems like something this doesn't need to guard against.

sgrif: My gut says crates.io is more likely to be down than GitHub except I ran the ops for crates for years and by the end nearly every time I got paged it was because the index couldn't update because GH was down >_<

phiggins: haha

phiggins: It feels like checking an env variable to allow overriding the source location is something we might want to support eventually, and that would handle the case where someone wants to build with no live internet connection.

Definitely not something necessary in the short term.

PR #326, opened 2021-06-14
Add Windows CI runner.

phiggins:


phiggins:
It passed but did not appear to do anything lol

phiggins:
The main ruby repo builds on windows on github actions so we could probably borrow what we need from there:

https://github.com/ruby/ruby/blob/master/.github/workflows/windows.yml
https://github.com/ruby/ruby/blob/master/.github/workflows/mingw.yml

phiggins:
Debugging this via CI has been obnoxious. I looked into whether there is a prebuilt VM available with the tools used to build this project, but the github actions team does not distribute the images they build (actions/virtual-environments#176) and I can't find anything else similar out there. I'm not really interested in getting a vanilla Windows vm and having to install rust, ruby, visual studio, etc.

 I'd happily debug the build if something like this were available, but it appears this is not the case, so I'm closing this.

phiggins:
Someone on twitter pointed me to this windows 10 image with visual studio from microsoft, which appears to be about 25% of what I wanted: https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/

PR #350, opened 2022-11-08
[bugfix] Bump bundled ruby to 2.7 branch to include fix for m1

bibstha:
the ruby submodule was pointing to an older version of ruby 2.7 which did not contain fix for M1 apple silicon.
The latest fixes have not yet been released as a patch version for ruby 2.7.X. I'm now using the default ruby 2.7 branch which works with rubyfmt without any issue.

With this patch, I can successfully build rubyfmt on M1 apple silicon (also tested in Mac OSX Ventura)

For example: https://github.com/ruby/ruby/pull/6297 is where ruby has the patch for building in M1 but is not in 2.7.6 ruby tag.

cc: @penelopezone @reese

### DIFF
--- a/librubyfmt/build.rs
+++ b/librubyfmt/build.rs
@@ -53,6 +53,7 @@ fn main() -> Output {
         .object(ruby_checkout_path.join(&ripper))
         .include(ruby_checkout_path.join("include"))
         .include(ruby_checkout_path.join(".ext/include/arm64-darwin21"))
+        .include(ruby_checkout_path.join(".ext/include/arm64-darwin22"))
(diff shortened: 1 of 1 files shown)


stepheneb:
Don't know when the next release for Ruby 2.7 is planned.

In case it's helpful for review here are all the changes between v2.7.6 and 2_7: https://github.com/ruby/ruby/compare/v2_7_6...ruby_2_7

reese (approved):
LGTM -- I don't think any of the changes to ruby in this bump should really affect parsing, so this should be totally fine. Thanks for the fix! 🎉

Issue #344, opened 2022-11-04
Rubymine instructions incorrect


shkm:
Screenshot shows arguments as `-i $FilePath$` but it should be `-i -- $FilePath$`. I'd make a PR but I fear the screenshot would be lower res and therefore uglier.

Alternatively, you could provide an XML file to import the watcher. I exported mine here:

```xml
<TaskOptions>
  <TaskOptions>
    <option name="arguments" value="-i -- $FilePath$" />
    <option name="checkSyntaxErrors" value="true" />
    <option name="description" />
    <option name="exitCodeBehavior" value="ERROR" />
    <option name="fileExtension" value="rb" />
    <option name="immediateSync" value="true" />
    <option name="name" value="rubyfmt" />
    <option name="output" value="$FilePath$" />
    <option name="outputFilters">
      <array />
    </option>
    <option name="outputFromStdout" value="false" />
    <option name="program" value="rubyfmt" />
    <option name="runOnExternalChanges" value="true" />
    <option name="scopeName" value="Project Files" />
    <option name="trackOnlyRoot" value="false" />
    <option name="workingDir" value="" />
    <envs />
  </TaskOptions>
</TaskOptions>
```

Thank you for your work!

fables-tales:
pull requests welcome!

reese:
Fixed in #353, thanks for the update!

PR #353, opened 2022-11-09
Update RubyMine information with watchers.xml

shkm:
See https://github.com/penelopezone/rubyfmt/issues/344


reese (approved):
👍  LGTM, thanks for the update!

PR #357, opened 2022-11-13
Document `.rubyfmtignore` and `.gitignore` behavior

### LINKED ISSUE #354: VSCode integration with --header-opt-out
Hi, I would like to use the opt-out argument to aid the adoption on a complex project.

However, it looks like the VSCode extension does not support additional arguments. Would it be possible to make --header-opt-out a default setting?

reese:
Resolves #354 

We added support for ignore-files some time ago, and I think some users will find them pretty useful, especially in bigger projects, so adding some docs to the README for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ Rubyfmt supports the following CLI invocations:
 * `rubyfmt --header-opt-in -- files or directories` to format files only with a `# rubyfmt: true` comment at the top of the file
 * `rubyfmt --header-opt-out -- files or directories` to skip formatting files with a `# rubyfmt: false` comment at the top of the file
 
+`rubyfmt` also supports ignoring files with a `.rubyfmtignore` file when present in the root of the working directory.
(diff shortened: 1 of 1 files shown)


matthewford:
Hi, I'm trying out using .rubyfmtignore with VS Code. However, it looks like the file is not being respected.

fables-tales (approved):

reese:
@matthewford is it respected if you run `rubyfmt` directly in the root directory?

matthewford:
@reese I suspect it is.

reese:
In that case, that sounds like something in the VSCode extension and should probably be filed as an issue there. I don't have a _ton_ of context on how that extension works though.

reese:
Ah it looks like it passes things through `stdin`, which is probably why this doesn't work. I had a thought that we should support something like `--stdin-filepath` that [prettier has](https://prettier.io/docs/en/options.html#file-path), which is basically "format these contents but treat them like they're at this file path to handle e.g. file ignores," but we don't have that currently (tracking in #365)

PR #358, opened 2022-11-14
Support unparenthesized Gemfile methods

### LINKED ISSUE #330: Add Gemfile `gem` method as special, un-parenthesised method
Similar to how RSpec's DSL methods stay un-parenthesised, I think adding parentheses to `gem` method calls in `Gemfile`s look weird. And perhaps weird enough to warrant a(nother) special case.

## Input file

```ruby
source "https://rubygems.org"
ruby "2.7.3"
gem "rspec"
```

## Rubyfmt's output

```ruby
source("https://rubygems.org")
ruby("2.7.3")
gem("rspec")
```

reese:
Resolves #330

This adds support for `gem`, `source`, `ruby`, and `group` as special unparenthesized methods. This logic is a _little_ duplicated since we separately handle methods with and without blocks (and some of these methods handle both), so I pulled other similar method names out as constants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -608,6 +608,7 @@ dependencies = [
  "backtrace",
  "cc",
  "jemallocator",
+ "lazy_static",
  "libc",
  "log",
  "ripper_deserialize",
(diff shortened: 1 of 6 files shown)


matthewford:
it would be good to also add rake tasks, and rails routes

reese:
Maybe, but I would treat those as out of scope for this PR (you'll probably want to open a separate ticket for that)

fables-tales (approved):

PR #369, opened 2022-11-22
Generically handle all method visibility modifiers

### LINKED ISSUE #363: Extra spacing between method decorators and definitions that use `module_function`
* Ruby version: 2.7.2
* Rubyfmt git sha: a65b40b694be57d189b89563061579929d1610ce

## Input file

```ruby
sig do
  void
end
module_function def my_func; end
```

## Rubyfmt's output

```ruby
sig do
  void
end

module_function def my_func
end
```

## This breaks because

For `private` et al., we don't add a blank line after the `end`, but we don't do that for `module_function`. We could add it to [this list](https://github.com/penelopezone/rubyfmt/blob/3a978ffe049adec2005b756597c802d5447eef61/librubyfmt/src/line_tokens.rs#L122-L126), but it might be better to do this for any methods that take a definition as the only arg.

reese:
Resolves #363

Previously we explicitly listed out all "visibility modifiers" like `private`, `private_class_method`, etc. to avoid adding extra newlines to them when rendering "method modifiers" like `sig`s. However, there will probably always be some we could be missing this way (right now, this includes things like `module_function` and `ruby2_keywords`), so this PR changes it such that we generically this for anything of the shape
```ruby
end
<whatever> def func
```
which seems more reasonable, especially if a user defined their own special method (like Sorbet's custom `package_private`).

### DIFF
--- a/librubyfmt/src/render_queue_writer.rs
+++ b/librubyfmt/src/render_queue_writer.rs
@@ -75,7 +75,17 @@ impl RenderQueueWriter {
                                 (ConcreteLineToken::Space, ConcreteLineToken::DefKeyword)
                             )
                         {
-                            accum.insert_blankline_from_end(4);
+                            // If we're here, the last few tokens must look like this:
+                            // | token             | index_from_end |
+                            // |  End              | 6              |
+                            // |. AfterCallChain   | 5              |
+                            // |  HardNewline      | 4              | <-- insert after this token
+                            // |. Indent           | 3 .            |
+                            // |  (ArbitraryToken) | 2              |
+                            // |  (ArbitraryToken) | 1              |
+                            // |  (ArbitraryToken) | 0              |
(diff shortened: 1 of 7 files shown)


librubyfmt/src/render_queue_writer.rs
```diff
@@ -63,16 +63,19 @@ impl RenderQueueWriter {
...
                     _ => {
                         if x.is_in_need_of_a_trailing_blankline()
-                            && !x.is_method_visibility_modifier()
+                            && !matches!(
+                                (maybe_space, maybe_def),
+                                (ConcreteLineToken::Space, ConcreteLineToken::DefKeyword)
+                            )
                         {
-                            accum.insert_trailing_blankline(BlanklineReason::ComesAfterEnd);
+                            accum.insert_blankline_from_end(4);
```
fables-tales: can we constantize/give a name to this 4?

fables-tales (approved):

PR #381, opened 2022-12-03
Handle arbitrary expression types in `aref_field` nodes

### LINKED ISSUE #380: Command calls in ArefFields fail to deserialize
* Ruby version: 2.7.3
* Rubyfmt git sha: homebrew release, see version check below

```
% rubyfmt --version                                            
rubyfmt-main 0.8.0-pre
```
```
% brew list rubyfmt         
/opt/homebrew/Cellar/rubyfmt/0.8.0/.crates.toml
/opt/homebrew/Cellar/rubyfmt/0.8.0/.crates2.json
/opt/homebrew/Cellar/rubyfmt/0.8.0/bin/rubyfmt
/opt/homebrew/Cellar/rubyfmt/0.8.0/bin/rubyfmt-main
```

## Input file

I can't share my source code at the moment as it doesn't belong to me, but I could change the names of everything to generic names and just share the structure of it if that would help.

~Here is a general skeleton of what it looks like (the real file is around 60 lines long):~

Edit 2022-12-03: (Updated code sample)
Here is the real structure of the original source code with a bunch of names changed to anonymize:
```ruby
# frozen_string_literal: true

module Fruit
  module Peach
    class Distribution
      attr_reader :baskets

      def initialize(baskets:)
        @baskets = baskets
        @modifiers = {}
      end

      def prioritized_baskets
        modify_baskets
        prune_baskets
        set_primary_basket
        baskets
      end

      private

        attr_reader :modifiers

        def modify_baskets
          baskets.each do |basket|
            baskets.contents.each do |item|
              modifier = find_modifier(item) || build_modifier(item)
              modifier.modify(baskets, item)
            end
          end
        end

        def build_modifier(item)
          modifiers[hash_item item] = Modifier.new(
            required_quantity: item.unit.required_quantity
          )
        end

        def find_modifier(item)
          modifiers[hash_item item]
        end

        def prune_baskets
          baskets.reject!(&:empty?)
        end

        def set_primary_basket
          baskets.first.is_primary = true if baskets.present? && baskets.none?(&:is_primary)
        end

        def hash_item(item)
          shipment = item.unit.shipment
          product  = item.unit.product

          return product.hash unless shipment

          product.hash ^ shipment.hash
        end
    end
  end
end
```

I found https://github.com/fables-tales/rubyfmt/blob/trunk/RELEASE_README.md and tried running the debug binary like so
```
% rubyfmt-debug -i -- <path-to-my-file>.rb 
zsh: command not found: rubyfmt-debug
```
I also tried installing the separate binary from Homebrew like so
```
% brew install rubyfmt-debug
...
Warning: No available formula with the name "rubyfmt-debug". Did you mean rubyfmt?
```

## Rubyfmt's output

```
% rubyfmt -i -- ./<path-to-my-file>.rb 
Error! source: ./<path-to-my-file>.rb

!!! Ruby Tree Deserialization Error !!!

Rubyfmt failed to correctly deserialize a tree from ripper. This is a bug that needs to be reported.
File a bug report at https://github.com/penelopezone/rubyfmt/issues/new.
Ideally you would include the full source code of the program you ran rubyfmt with.
If you can't do that for some reason, the best thing you can do is rerun rubyfmt on this program 
with the debug binary with `2>log_file` on the end and then send us the log file that gets generated.
```

## This breaks because

`rubyfmt` should be able to execute against *.rb files without throwing a `Ruby Tree Deserialization Error`. 

### Question

How does one run the `rubyfmt-debug` binary - is it available from binary that is built from source, and not the binary that is installed via Homebrew?

reese:
Resolves #380 

`aref_field` nodes (e.g. `foo[:bar] = :baz`) can really handle _any_ expression in the index position, and in reality they can handle any _number_ of items for users overriding `#[]=`, so we should handle an arbitrary expression list here.

### DIFF
--- a/fixtures/small/aref_field_expected.rb
+++ b/fixtures/small/aref_field_expected.rb
@@ -20,3 +20,22 @@ a[begin
 a[] = ""
 # Users can override #[]= to have multiple args
 a[1, 2] = ""
+
+a[puts(a)]
+a[puts(a)]
+a[()]
+
+a[class A
+    def foo
+      puts(a)
+    end
+end]
+a[begin
+    ""
+  rescue StandardException
+    ""
+  end]
+# Users can override #[] to not have args
+a[]
+# Users can override #[] to have multiple args
+a[1, 2]
--- a/fixtures/small/aref_field_expected.rb
+++ b/fixtures/small/aref_field_expected.rb
@@ -1,3 +1,22 @@
 # As above
 a[0] = b
 # So below
+
+a[puts(a)] = ""
+a[puts(a)] = ""
+a[()] = ""
+
+a[class A
+    def foo
+      puts(a)
+    end
+end] = ""
+a[begin
+    ""
+  rescue StandardException
+    ""
+  end] = ""
+# Users can override #[]= to not have args
+a[] = ""
+# Users can override #[]= to have multiple args
+a[1, 2] = ""
--- a/fixtures/small/aref_field_expected.rb
+++ b/fixtures/small/aref_field_expected.rb
@@ -5,17 +5,20 @@ a[0] = b
 a[puts(a)] = ""
 a[puts(a)] = ""
 a[()] = ""
-
-a[class A
+a[
+  class A
     def foo
       puts(a)
     end
-end] = ""
-a[begin
+  end
+] = ""
+a[
+  begin
     ""
   rescue StandardException
     ""
-  end] = ""
+  end
+] = ""
 # Users can override #[]= to not have args
 a[] = ""
 # Users can override #[]= to have multiple args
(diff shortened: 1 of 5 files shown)


fixtures/small/aref_field_expected.rb
```diff
@@ -1,3 +1,47 @@
...
+    end
+  end
+] = ""
+a[
+  begin
+    ""
+  rescue StandardException
+    ""
+  end
+] = ""
```
reese, line 21: I'm gonna call this a "reasonable" rendering? I mean, users writing this is already pretty surprising, so I think this is the most consistent way to handle it rather than special-casing aref fields to inline these or something like that 🤷‍♂️

librubyfmt/src/render_targets.rs
```diff
@@ -144,7 +144,9 @@ impl AbstractTokenTarget for BreakableEntry {
     }
 
     fn is_multiline(&self) -> bool {
-        self.line_numbers.len() > 1 || self.any_collapsing_newline_has_heredoc_content()
+        self.line_numbers.len() > 1
+            || self.any_collapsing_newline_has_heredoc_content()
+            || self.contains_hard_newline()
```
reese, line 149: I also think this was a latent bug we've had around forever -- some expressions are essentially _always_ multiline (e.g. classes), so breakables that have these in them should always render multiline.

fables-tales (approved):

Issue #359, opened 2022-11-14
User-multilined method chains and binaries won't stay multilined


reese:
* Rubyfmt git sha: ed99cc4586a908c97f8b19ed78801342f7aa8512

## Input file

```ruby
foo
  .bar
  .baz

foo &&
  bar ||
  baz
```

## Rubyfmt's output

```ruby
foo.bar.baz

foo && bar || baz
```

## This breaks because

For other "breakable" constructs like method params, array literals, etc., we use the multiline form if the user originally multilined it, but we don't do this for method chains and binaries since they have "special" multilining implementations (e.g. they don't use `breakable_of` internally). For consistency's sake, we should multiline user-multilined constructs for these.

PR #382, opened 2022-12-03
Properly multiline user-multilined call chains

reese:
Related to #359 

In general, `rubyfmt` respects (within limits) user choices about multilining expressions, but we don't currently do this for multiline expressions. This has a few poor side effects: (1) users often _want_ to multiline to make e.g. adding comments for specific sections of a call chain easier, and (2) there end up being some (IMO) awkward formatting of chains like the following:
```ruby
MyClass.call(
  [],
  ""
).do_some_other_stuff!
```
which is inconsistent with `rubyfmt`'s normal call chain styling.

This PR implements multilining for user-multilined expressions. It does this by checking the line location of the `DotTypeOrOp` items in the call chain and multilining if they're on multiple lines. This _might_ be a slightly imperfect if the user writes something like this

```ruby
# both format as `foo.bar.baz`
foo
.bar.baz

foo.bar.
baz
```

but IMO this is a reasonable implementation tradeoff. We _could_ alternatively check every expression's start/end, but this ends up requiring a _lot_ more trickery and is more prone to inconsistency.

### DIFF
--- a/fixtures/small/method_chains_expected.rb
+++ b/fixtures/small/method_chains_expected.rb
@@ -15,6 +15,28 @@ returns_array
     b: ""
   )
 
+foo.bar.baz
+
+foo
+  .bar
+  .baz
+
+# If they're all on the same line but different from
+# the first receiver, consider that "on one line"
+foo.bar.baz
+
+foo::bar&.nil?
+
+foo
+  ::bar
+  &.nil?
+  ::klass
+  .true?
+
+Class
+  &.new
+  .call!
+
 def example
   things
     .map do |thing|
--- a/fixtures/small/method_chains_expected.rb
+++ b/fixtures/small/method_chains_expected.rb
@@ -27,10 +27,8 @@ foo.bar.baz
 
 foo::bar&.nil?
 
-foo
-  ::bar
-  &.nil?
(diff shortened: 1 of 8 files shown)


fixtures/small/method_chains_expected.rb
```diff
@@ -15,6 +15,26 @@
...
+
+# If they're all on the same line but different from
+# the first receiver, consider that "on one line"
+foo.bar.baz
+
+foo::bar&.nil?
+
+foo::bar
+  &.nil?::klass
+  .true?
```
reese, line 32: This is `blursed` but technically legal -- I handled it by just multilining _where possible_, but open to other thoughts on how we should handle `::` calls.

fables-tales: I hate you

fables-tales (approved):

fixtures/small/method_chains_expected.rb
fables-tales: (j/k obvi)

reese: I think pretty much every fixture I write should just have `# im so sorry` at the top of them

Issue #386, opened 2022-12-14
CLI incorrectly reports a new version being available on the most recent release


reese:
* Ruby version: 2.7.4
* Rubyfmt git sha: `0.8.0-pre` (ed99cc4586a908c97f8b19ed78801342f7aa8512)

This occurs on any input:

```bash
% echo "class Foo; end" | rubyfmt
class Foo
end
A new version of rubyfmt is available at https://github.com/penelopezone/rubyfmt/releases/tag/v0.8.0
% rubyfmt --version
rubyfmt-main 0.8.0-pre
```

## This breaks because

I would expect this not to tell me a new version is available if I'm on the latest version. This might be related to the `-pre` tag, but I haven't looked enough at how this update check works to know if that matters.

reese:
Ah I actually know why this is -- when we released v0.8.0, in the Cargo.toml we actually marked it as a prelease (`v0.8.0-pre`). We _also_ did this in the GitHub release originally, but [brew doesn't let you release something marked as a prerelease on GitHub](https://github.com/fables-tales/rubyfmt/issues/349#issuecomment-1321012483), so we made it no longer a prerelease, so the _GitHub_ version marker _is_ technically different (and this update check is done by querying the latest GH release).

**TL;DR** -- I don't really think there's anything we should do here other than we should no longer be marking these as prereleases internally, because (a) having it on brew means it necessarily _can't_ be a prerelease and (b) this is already supporting millions of lines of code in production daily, so I think it's fair that it's no longer a prerelease anyways.

PR #387, opened 2022-12-15
Delete version update check

reese:
Related to #386

This version check is neat but ultimately pretty error-prone (and I imagine most users tune it out, given that it's broken on the most recent release and nobody's said anything about it 🤷 ). The version in the `Cargo.toml` and on GitHub are not _strictly_ equivalent, so that makes the value of this check far more  prone to mistakes and might be annoying/confusing when we mess it up.


build.rs
```diff
@@ -3,6 +3,7 @@ fn main() {
     #[cfg(target_os = "macos")]
     {
         println!("cargo:rustc-link-lib=framework=foundation");
+        println!("cargo:rustc-link-lib=framework=security");
```
reese, line 6: Had to add this to get this working on my M1 -- not sure why it seemed to work for others, although I don't know if most folks use the Makefile (which has this [here](https://github.com/fables-tales/rubyfmt/blob/9ed05fcf5c80c636110fe3430ae9a84ecc425aee/Makefile#L5)).

fables-tales (approved):

PR #393, opened 2022-12-19
Fix rendering of single-item massigns for rest params 

### LINKED ISSUE #392: rubyfmt produces invalid ruby on wayward splats
* Ruby version: 2.7.6
* Rubyfmt git sha: [v0.8.1](https://github.com/fables-tales/rubyfmt/releases/tag/v0.8.1)

## Input file

```ruby
a = 1
b = 2
c = 3
*nums = [a, b, c]

p nums
```

N.B. - I don't know why we have that asterisk there, or what it's meant to achieve, but it is valid ruby and is the only bug I noticed when running rubyfmt on our ~400k lines of Ruby at work 😲 

## Rubyfmt's output

```
a = 1
b = 2
c = 3
*nums, = [a, b, c]

p(nums)
```

## This breaks because

This is not valid ruby.

reese:
Resolves #392 

For massigns, we have special handling for the case where the massign is only one item, e.g. `foo, = [1, 2]`, since that comma produces different behavior at runtime. However, Ruby also considers `*foo = [1, 2]` an massign, even though there's only one item and no comma. Ruby _also_ doesn't consider `*foo, = [1, 2]` valid syntax, so we shouldn't render that final comma when the only item in an massign is a rest param.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -642,7 +642,7 @@ dependencies = [
 
 [[package]]
 name = "rubyfmt-main"
-version = "0.8.0-pre"
+version = "0.8.1"
 dependencies = [
  "atty",
  "clap",
--- a/fixtures/small/massign_actual.rb
+++ b/fixtures/small/massign_actual.rb
@@ -1,5 +1,6 @@
 class Foo
   def bees
     a, b = [1, 2]
+    *nums = [a, b, c]
   end
 end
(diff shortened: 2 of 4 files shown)


Cargo.lock
```diff
@@ -642,7 +642,7 @@ dependencies = [
 
 [[package]]
 name = "rubyfmt-main"
-version = "0.8.0-pre"
+version = "0.8.1"
```
reese, line 645: I think the lockfile didn't get updated during the last version bump, so I included it here 🤷‍♂️

fables-tales (approved):

Issue #399, opened 2022-12-30
[Tracking] Unsupported features (as of Ruby 3.2)


reese:
`rubyfmt` currently doesn't support many of the major syntax additions of the last year. We historically targeted 2.7, but with the 2.7 EOL looming and the addition of so many features in 3.0-3.2, we should start looking at supporting the new features. This ticket is mostly a tracking issue to have visibility on the progress here.

- [x] [Pattern matching](https://rubyreferences.github.io/rubychanges/3.0.html#pattern-matching) (existed in 2.7 but stabilized in 3)
  - [x] [2.7 style pattern matching](https://rubyreferences.github.io/rubychanges/2.7.html#pattern-matching)
  - [x] [Rightward assignment](https://rubyreferences.github.io/rubychanges/3.0.html#one-line-pattern-matching-with-)
  - [x] [`in` matching](https://rubyreferences.github.io/rubychanges/3.0.html#in-as-a-truefalse-check)
  - [x] [Pin operator](https://rubyreferences.github.io/rubychanges/3.1.html#expressions-and-non-local-variables-allowed-in-pin-operator-)
- [x] [Endless methods](https://rubyreferences.github.io/rubychanges/3.0.html#endless-method-definition)
  - Implemented in #405 
- [x] [Hash literal value omissions](https://rubyreferences.github.io/rubychanges/3.1.html#values-in-hash-literals-and-keyword-arguments-can-be-omitted)
  - Implemented in #404
- [x] [Anonymous block args](https://rubyreferences.github.io/rubychanges/3.1.html#anonymous-block-argument)
  - Implemented in #471

rattrayalex:
Hi @reese! I'd love to fund work on this if you're open to sponsorship / contract work?

reese:
Hi @rattrayalex, thanks for reaching out! I'm currently on a bit of a break from more formal work arrangements and so probably won't have the capacity for sponsorship or contract work at the moment. If anyone else would has interest in working on this, I'd be more than happy to give pointers in the codebase for how any of this could be done, and #456 does a pretty good chunk of the heavy lifting for getting pattern matching over the line anyways, so I hope that with that it's actually not a _huge_ amount of work.

(It may also be relevant to check with @fables-tales to see if any of the folks currently working at Stripe have this work on their radar, since I know there were at least very-early-stage discussions about pattern matching a few months ago.)

reese:
[...]

PR #400, opened 2022-12-30
Bump Ruby checkout to 3.0

reese:
There are several new features in newer versions of Ruby that we should eventually add support for, so this is the first of several PRs to incrementally get the checkout up to date. I'll probably jump straight to 3.2 after this, but I had to make some revisions just to get 3.0 working so I figured I'd do them separately.

This will unblock much of the work on #399, with the rest unblocked once we get to 3.2.

(I also ran this over the Stripe codebase internally and it didn't seem to make any different changes from master and didn't crash, so I think this should be safe.)


librubyfmt/rubyfmt_lib.rb
```diff
@@ -146,25 +146,25 @@ def escape_percent_array_paren_content(part, pattern)
...
-            /(?<!\\)(?:\\\\)*(?:\\#{Regexp.escape(delim_start)}|[\[\]])/
-          end
-          if part[0].is_a?(Array)
-            part.each do |sub_part|
-              escape_percent_array_paren_content(sub_part, pattern)
-            end
-          else
-            escape_percent_array_paren_content(part, pattern)
+
+      unless delim.end_with?('[')
```
reese, line 150: This change is due to https://github.com/ruby/ruby/pull/3281, which moved `tap` from `Object` to `Kernel` in Ruby 3. In most cases that's fine, but since we're kinda hacking together Ruby bootstrapping here, `Kernel` hasn't been loaded yet, so `tap` doesn't exist. I also don't think we really _needed_ it here, so I just removed it.

librubyfmt/src/lib.rs
```diff
@@ -193,6 +194,7 @@ unsafe fn load_ripper() -> Result<(), ()> {
         "../ruby_checkout/ext/ripper/lib/ripper/sexp.rb"
     ))?;
 
+    rb_gc_disable();
```
reese, line 197: Totally open to suggestions on a better way to do this since I couldn't _really_ figure out why `GC.disable` seemed to be undefined after I bumped to 3, but I ended up just calling the underlying VM implementation here instead to get ~the same result.

fixtures/small/nested_destructuring_actual.rb
```diff
@@ -1 +1 @@
-_1, (_2, _3) = []
+_a, (_b, _c) = []
```
reese, line 1: Related to https://github.com/ruby/ruby/pull/3163 -- in Ruby 3, assigning to numbered param is now a syntax error. This _is_ technically a backwards-incompatible change for rubyfmt, but IMO I think that's fine.

fables-tales (approved):

PR #403, opened 2023-01-03
Format all file paths when not traversing directories

### LINKED ISSUE #396: Rubyfmt ignores files without the .rb extension (for example `.rake` files)
* Ruby version: 2.7.0p0
* Rubyfmt git sha: 266b47b (0.8.1)

## Input file

```ruby
desc "Do thing"
task :do_thing do
  puts "Doing things"
end
```

## Rubyfmt's output
rubyfmt -i  task.rake 
```
[DEBUG] (1) rubyfmt: logger works
```

## This breaks because

rubyfmt ignores the file because it has a `.rake` extension, even when prompted to format it explicitly.
Renaming the file to `task.rb` works around the issue.

I believe this is https://github.com/fables-tales/rubyfmt/blob/9a24d922686e42df014095b3e5d12fcb9e4146d6/src/main.rs#L302-L303

See also #189 & #210

To me there are two related but slightly different issues:
* rubyfmt picking up files by default when running on a directory − we can discuss whether it should pick more than `.rb` (as #210 did) but I think it’s a sensible default
* rubyfmt refusing to format my `task.rake` file even though I specifically invoke `rubyfmt -i  task.rake`

reese:
Fixes #396 

If a user passes an explicit file path, we should always format it, regardless of its extension. This requires a bit of a messy implementation to be able to know which files are passed on the command line, so I implemented it by separately running the file walker on just the files and then just the directories.

In the future, we could allow users to pass custom file paths, but I don't think that's strictly required for sake of #396.

### DIFF
--- a/script/tests/test_cli_interface.sh
+++ b/script/tests/test_cli_interface.sh
@@ -654,6 +654,23 @@ DIFF
     )
 }
 
+test_formats_non_rb_files() {
+    (
+    cd "$(mktemp -d)"
+    echo "a 1, 2, 3" > test.rake
+
+    f_rubyfmt -i -- test.rake
+
+    cat > test_expected.rake <<- DIFF
+a(1, 2, 3)
+DIFF
+
+    cat test.rake
+
+    diff_files o test_expected.rake test.rake
+    )
+}
+
 test_simple_stdout
 test_i_flag
 test_i_flag_no_changes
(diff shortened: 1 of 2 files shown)


fables-tales (approved):

Sinjo:
@reese Any chance a release could be cut for this one? Ran into it today on my first time using `rubyfmt` (I have a script in the `bin` folder of my project that I couldn't get it to operate on).

reese:
It's being worked on, but I don't have permissions to cut releases, I believe @fables-tales is on it.

Sinjo:
Thank you 💖

PR #407, opened 2023-01-04
Transform multiline brace blocks to do/end blocks

reese:
It's a pretty standard convention to use `{ }` for single-line blocks and `do`/`end` for multiline blocks, and for the sake of consistency, we should also enforce that in rubyfmt.

(Note that while there _are_ edge cases in terms of precedence differences between the two block types, this only happens in the case of unparenthesized arguments that use brace blocks, and since transforming them to `do/end` makes them multiline, rubyfmt will always wrap them in parentheses, which avoids the precedence issues. Because of that, this should be a safe transformation.)

### DIFF
--- a/fixtures/small/block_local_variables_expected.rb
+++ b/fixtures/small/block_local_variables_expected.rb
@@ -1,10 +1,10 @@
 y = 12
-lambda { |x, ;y|
+lambda do |x, ;y|
   puts(x)
   puts(y)
   y = 19
   puts(y)
-}
+end
   .call(17)
 puts(y)
 
--- a/fixtures/small/blocks_with_only_comments_expected.rb
+++ b/fixtures/small/blocks_with_only_comments_expected.rb
@@ -15,7 +15,7 @@ def foo
   end
 
   # More comments!!
-  commented_brace_block {
+  commented_brace_block do
     # Such plainness of the pre-baroque
     # Hardly involves the eye, until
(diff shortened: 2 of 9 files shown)


fables-tales:
I like multiline brace blocks

reese:
I personally don't mind them, but I think the opinion from users (and the opinion I thought we landed on when we discussed this a few months back) was that historically we've operated on the rationale that it was best to pick one thing that worked everywhere and always use that, and allowing both braces and do/end broke that consistency. The other context was (1) the ideal way would be brace blocks for blocks that return things (e.g. `map` blocks) and do/end for blocks used for side effects (e.g. `each` blocks), but (2) that wasn't possible in rubyfmt so we were planning on going this way.

I think last time we talked about this it was another one of those "make fable really sad but at least it's consistent" things.

shkm:
Random Ruby person here sliding in your PR. For what it's worth, I'd rather like this. But...

I just wanted to point out that there are some DSLs which promote use of multi-line brace blocks. The only example I can think of, off the top of my head, is [Arbre](https://activeadmin.github.io/arbre/).

reese:
There will _always_ be some degree of "other people do it differently," but that's sort of the entire point of having an opinionated formatter: it doesn't really _matter_ what other tools do. I think the question here is whether or not that's the style rubyfmt should go with, not a question of whether or not it satisfies every possible style used by other tools.

PR #410, opened 2023-02-12
Allow `test` to be an unparenthesized method

### LINKED ISSUE #383: Add `test` to the list of methods that do not get parentheses
* Ruby version: 2.7.2
* Rubyfmt git sha: 9ed05fcf5c80c636110fe3430ae9a84ecc425aee

## Input file

```ruby
test "this is a test in rails" do
  assert(true)
end
```

## Rubyfmt's output

```
test("this is a test in rails") do
  assert(true)
end
```

## This breaks because

There was an old issue: https://github.com/fables-tales/rubyfmt/issues/54 that mentions a bunch of keywords that shouldn't have `()` after the method call. However `test` has not been included in the list in `rubyfmt`.

`test` is like `describe` or `it` keywords that Rails uses to define test cases and is widely used. Therefore this should be added too.

Screenshot 2022-12-05 at 10 17 44 PM

reese:
Closes #383 

`test` works just like `it` and `describe` in Rails tests, so I think it's fine to let it not use parens just like the other test methods.

### DIFF
--- a/fixtures/small/rspec_its_actual.rb
+++ b/fixtures/small/rspec_its_actual.rb
@@ -39,6 +39,10 @@ describe "writing some really really long test name",
   opinion: "please don't name your DSL 'describe', but a little too late now I guess" do
 end
 
+test "some really good test name" do
+  assert(true)
+end
+
 RSpec.describe "bees" do
 end
 
--- a/fixtures/small/rspec_its_expected.rb
+++ b/fixtures/small/rspec_its_expected.rb
@@ -43,6 +43,10 @@ describe(
 ) do
 end
 
+test "some really good test name" do
+  assert(true)
+end
+
 RSpec.describe "bees" do
 end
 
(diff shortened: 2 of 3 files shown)


fables-tales:
what if we just did a sin and forced all these methods to be parenthesized?

bibstha:
Would love to see it get merged. 

> what if we just did a sin and forced all these methods to be parenthesized?

@fables-tales Rails and a bunch of ruby test frameworks use these without `()` today. It makes it readable, plus that is the accepted norm. Sometimes the rails test runner doesn't pickup the line-numbering when defined with `test("my test case") do; end` when running a single test case.

pjg:
Could we add `context` to this list as well? It's so weird seeing `rubyfmt` adding braces to all my `context 'xxx' do` methods, while leaving `it` and `describe` without them.

pjg:
And similarly all other RSpec "helper" methods should also be left without parenthesis: `to`, `not_to`, `to_not`, `be`, `eql`, `include` and so on.

Observing `expect(country).to eql 'Ukraine'` be transformed into `expect(country).to(eql('Ukraine'))` is just not great as the existing convention, which I believe to be the norm, is to omit parentheses in such situations.

I wish this was configurable in some way for the end user, though I understand the goal of the project being configuration-less.

PR #415, opened 2023-02-22
Do not elide parens when special-cased names are used in call chains

### LINKED ISSUE #414: Dropping brackets around string arguments in chained method calls
* Ruby version: 3.1.3
* Rubyfmt git sha: 266b47b4666106e579626b1823100ccedf5cbbc3

## Input file

```ruby
scope :with_ticket_types, -> { group('shops.id').joins('LEFT JOIN foos ON foos.bar_id 
= bars.id').select('foos.*, COUNT(bars.id) count') }
```

## Rubyfmt's output

```
scope(
  :with_ticket_types,
  -> { group "shops.id".joins "LEFT JOIN foos ON foos.bar_id = bars.id".select("foos.*, COUNT(bars.id) count") }
)

```

## This breaks because

There are no brackets over `group` and `join` calls so that invalid syntax is being produced - `joins` & `select` are called on a string instances.

reese:
Closes #414 

In some cases, we special-cased certain "reserved words" for common usages like Gemfiles and tests. These should only ever be used outside of proper "call chains," i.e. they should only really be called directly with arguments and not chained to other methods. However, if someone _happens_ to use a method with the same name in a call chain, we should correctly wrap it in parens.

### DIFF
--- a/fixtures/small/gemfile_actual.rb
+++ b/fixtures/small/gemfile_actual.rb
@@ -10,3 +10,9 @@ end
 group :test do
   gem 'rspec'
 end
+
+# Usages of gemfile methods but in call chains
+
+scope :with_ticket_types, -> { group('shops.id').joins('LEFT JOIN foos ON foos.bar_id
+= bars.id').select('foos.*, COUNT(bars.id) count') }
+
--- a/fixtures/small/gemfile_expected.rb
+++ b/fixtures/small/gemfile_expected.rb
@@ -10,3 +10,15 @@ end
 group :test do
   gem "rspec"
 end
+
+# Usages of gemfile methods but in call chains
+
+scope(
+  :with_ticket_types,
+  -> {
+    group("shops.id")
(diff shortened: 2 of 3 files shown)


fables-tales (approved):

bquorning:
I just ran into this issue again, this time on another repo than when I opened #418 😅 When do you plan on cutting a new release?

bquorning:
@reese Now I ran into the same issue once more. Is there a new release in the works?

reese:
Yes, I've asked for us to cut a new release as well, and it should be in the works, although I don't know if there's a specific planned release date at the moment (I'm not personally able to cut releases).

bquorning:
v0.9.5 was released last week without much fanfare (46e08ebde71f5e816bdde0065c5100f33ee638a7). Now I’m just waiting for [the Homebrew formula](https://github.com/Homebrew/homebrew-core/pull/137470) to get updated too.

PR #431, opened 2023-06-13
remove `ureq` from dependencies

froydnj:
`ureq` was required in the distant past, but the code that depended on it was removed.  Eliminating `ureq` from the build system entirely makes things compile somewhat faster and lets us deal with fewer dependencies from the Rust side.

cc @ilyailya


reese (approved):
Love it

PR #432, opened 2023-06-14
build just the `main` target to avoid extraneous work

froydnj:
Per the tin.  This avoids needing to run rdoc over everything, among other things.


froydnj:
r? @reese 

Compared to the previous PR, this cuts test times for the Linux jobs in half, assuming the numbers are relatively stable.

reese (approved):

PR #435, opened 2023-06-14
extract ruby configuration for determining link libs

froydnj:
We currently statically determine whether `libz` and `libcrypt` should be linked into the `rubyfmt` binary.  A better approach (since Ruby can successfully be compiled without these libraries present, at the cost of functionality that `rubyfmt` doesn't use) is to query whether Ruby found these libraries on the system and output the necessary cargo directives based on that information.


froydnj:
r? @reese 

I think this + a similar patch for the main `rubyfmt` binary will make it so we can cross-compile to aarch64-linux-gnu without having to worry about the details of what libraries are available in the cross-compilation environment.  cc @ilyailya

reese (approved):

PR #436, opened 2023-06-14
add aarch64 cross-compilation bits to CI and release

froydnj:
This PR adds an aarch64-linux cross compilation job to the normal set of CI jobs.  The release process is also modified to run an aarch64-linux cross compilation job and package up the resulting bits, so that we are now producing three binary tarballs:

* x86-64 linux
* x86-64 mac
* aarch64 linux

The regular CI job is obviously tested in this PR; I don't know exactly how to test the release job.

None of the aarch64 artifacts are currently tested.  I believe that doing so is within the realm of possibility, but involves another layer of complexity that I would prefer to postpone until later.  I have a high degree of confidence that if the x86-64 linux binaries pass their tests, then the aarch64 linux binaries will run without issues.

cc @ilyailya

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, ubuntu-20.04, macos-latest]
+        include:
+          - os: ubuntu-latest
+            target: native
+          - os: ubuntu-20.04
+            target: native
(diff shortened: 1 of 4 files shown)


reese (approved):
LGTM -- you could in theory hack in some things if you wanted to manually test the release job, but I think it's fine for now since it's basically the same changes as the CI job (and we haven't even used the release action yet, so it's fine if we need to debug along the way)

PR #443, opened 2023-08-16
Model call chains as an `AbstractTokenTarget`

reese:
Resolves #397

(Apologies in advance for the large PR, but this a pretty foundational change that required quite a bit of boilerplate!)

This PR implements call chains as a new `AbstractTokenTarget`: `BreakableCallChainEntry`. This means that multilining for call chains -- just like for existing breakables -- will be decided on-the-fly during rendering, where we have more contextual information and can better enforce line length.

This is done with a combination of the normal breakable machinery and some additional "magic tokens." Specifically, this adds tokens at the beginning and end of the indent-able portion of the call chain, and then during rendering, we maintain a count of how many nested call chains we're in and adjust indentation accordingly. This isn't exactly the _cleanest_ implementation, but given the complexity in use-cases for call chains, it was the best I could muster in a way that functioned how users would generally expect.

This is a fairly large PR, but there's primarily three main portions which each are about a third of the total diff each: fixtures, the boilerplate in `ripper_tree_types` to support getting expression starting lines, and finally the main formatting changes. I've done my best to comment these areas appropriately, but it's helpful to add further explanation there, I'm more than happy to add more.

### DIFF
--- a/fixtures/small/heredocs_expected.rb
+++ b/fixtures/small/heredocs_expected.rb
@@ -82,6 +82,15 @@ this_one_is
     MYHEREDOC
   end
 
+def foo
+  "#{stuff.each do
+    <<~MESSAGE
+      #{message.text}
+    MESSAGE
+      .strip
+  end.join("\n\n")}"
+end
+
 puts(a)
 puts(b)
 foo
--- a/fixtures/small/heredocs_expected.rb
+++ b/fixtures/small/heredocs_expected.rb
@@ -69,6 +69,19 @@ add_offense(
   rubocop.smash(bad_thing)
 end
 
+this_one_is
+  .in_a_call_chain {
+    # some stuff
+  }
+  .each do
+    <<~MYHEREDOC
+      Words are pale shadows of forgotten names.
+      As names have power, words have power.
+      Words can light fires in the minds of men.
+      Words can wring tears from the hardest hearts.
+    MYHEREDOC
+  end
+
 puts(a)
 puts(b)
 foo
--- a/fixtures/small/single_line_method_call_chain_expected.rb
+++ b/fixtures/small/single_line_method_call_chain_expected.rb
@@ -1,3 +1,7 @@
 def blorp
   method_call(arg_a, arg_b).chained_call
 end
+
+This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain.build(
+  reason: Because::ThisCouldCauseUsSome::SeriousProblemsIfWeTriedToMultilineIt::EventThoughItActuallyDoesExtendBeyondTheLimit
+)
--- a/fixtures/small/single_line_method_call_chain_expected.rb
+++ b/fixtures/small/single_line_method_call_chain_expected.rb
@@ -2,9 +2,10 @@ def blorp
   method_call(arg_a, arg_b).chained_call
 end
 
-This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain.build(
-  reason: Because::ThisCouldCauseUsSome::SeriousProblemsIfWeTriedToMultilineIt::EventThoughItActuallyDoesExtendBeyondTheLimit
-)
+This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain
+  .build(
+    reason: Because::ThisCouldCauseUsSome::SeriousProblemsIfWeTriedToMultilineIt::EventThoughItActuallyDoesExtendBeyondTheLimit
+  )
 
 this.surprisingly.can_break
 # due to this comment being here!
--- a/fixtures/small/single_line_method_call_chain_expected.rb
+++ b/fixtures/small/single_line_method_call_chain_expected.rb
@@ -6,5 +6,9 @@ This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain.build(
   reason: Because::ThisCouldCauseUsSome::SeriousProblemsIfWeTriedToMultilineIt::EventThoughItActuallyDoesExtendBeyondTheLimit
 )
 
+This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain.build(
+  reason: Because::IPutThisOnMultipleLines
+)
+
 this.surprisingly.can_break
 # due to this comment being here!
--- a/fixtures/small/single_line_method_call_chain_expected.rb
+++ b/fixtures/small/single_line_method_call_chain_expected.rb
@@ -13,3 +13,20 @@ This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain
 
 this.surprisingly.can_break
 # due to this comment being here!
+
+# If there's a super long comment here that goes over the line limit weeeeeeeeeeeeeeeeeeee look at this one go
+this.wont.break
+
+# And there's a comment here
+this
+  .is_really_long
+  .and_will_go_beyond_the_maximum_line_length_and_break
+  .across_multiple_lines
+  .and_it_has_a_comment_after_it!
+
+this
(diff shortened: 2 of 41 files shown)


fixtures/small/heredocs_expected.rb
```diff
@@ -69,6 +69,19 @@ class Foo
   rubocop.smash(bad_thing)
 end
 
+this_one_is
+  .in_a_call_chain {
+    # some stuff
+  }
+  .each do
+    <<~MYHEREDOC
```
fables-tales, line 77: crimes

reese: on the downside, i'm currently still debugging even more crimes

librubyfmt/src/format.rs
```diff
@@ -2807,287 +2744,155 @@ fn can_elide_parens_for_reserved_names(cc: &[CallChainElement]) -> bool {
...
+                        if let Some(end_line) = end_line {
+                            // If we're rendering a single-line chain, force a reset so
+                            // that comments end up at the current indentation level
                             ps.reset_space_count();
+                            ps.wind_dumping_comments_until_line(end_line);
                         }
-                        ps.wind_dumping_comments_until_line(end_line);
                     }
+                } else if is_last_call_args && last_call_use_parens.unwrap_or(false) {
+                    ps.emit_open_paren();
```
fable-stripe: should this be using a delimiters?

librubyfmt/src/format.rs
```diff
@@ -1701,14 +1642,10 @@ pub fn format_heredoc_string_literal(
         Box::new(|ps| {
             let heredoc_type = (hd.1).0;
             let heredoc_symbol = (hd.1).1;
-            ps.emit_heredoc_start(heredoc_type.clone(), heredoc_symbol.clone());
+            let kind = HeredocKind::from_string(&heredoc_type);
+            ps.emit_heredoc_start(heredoc_type, heredoc_symbol.clone(), kind);
```
fable-stripe: if we're modifying emit_heredoc_start to take a HeredocKind is there a reason we still need to pass `heredoc_type`?

librubyfmt/src/format.rs
```diff
@@ -2807,287 +2744,155 @@ fn can_elide_parens_for_reserved_names(cc: &[CallChainElement]) -> bool {
...
             CallChainElement::Paren(p) => format_paren(ps, p),
             CallChainElement::IdentOrOpOrKeywordOrConst(i) => {
                 let ident = i.into_ident();
-                element_is_super_keyword = ident.1 == "super";
+                element_must_use_parens = ident.1 == "super";
 
-                format_ident(ps, ident)
+                if ident.1 == ".()" {
+                    ps.emit_ident(".".to_string());
+                    element_must_use_parens = true;
```
fable-stripe: the structure of `element_must_use_parens` here feels weird to me, but maybe this is the best way to do it?

librubyfmt/src/format.rs
```diff
@@ -2807,287 +2744,155 @@ fn can_elide_parens_for_reserved_names(cc: &[CallChainElement]) -> bool {
...
+                } else {
+                    format_ident(ps, ident);
+                }
+                ps.shift_comments();
             }
             CallChainElement::Block(b) => {
                 ps.emit_space();
                 format_block(ps, b)
```
fable-stripe, line 2806: are we missing a shift_comments call here?

librubyfmt/src/format.rs
```diff
@@ -2807,287 +2744,155 @@ fn can_elide_parens_for_reserved_names(cc: &[CallChainElement]) -> bool {
...
-                        Box::new(|ps| {
-                            format_list_like_thing(ps, aas, end_line, false);
-                        }),
-                    );
-                    if let Some(end_line) = end_line {
-                        // If we're rendering a single-line chain, force a reset so
-                        // that comments end up at the current indentation level
-                        if !render_multiline_chain {
+                        if let ArgsAddStarOrExpressionListOrArgsForward::ExpressionList(mut el) =
+                            aas.clone()
```
fable-stripe: is the clone call here to prevent the `pop` from modifying? If so can we add a comment?

librubyfmt/src/render_targets.rs
```diff
@@ -51,27 +52,60 @@ pub trait AbstractTokenTarget: std::fmt::Debug {
...
-    context: FormattingContext,
+    context: Vec<FormattingContext>,
 }
 
 impl AbstractTokenTarget for BreakableEntry {
     fn to_breakable_entry(self: Box<Self>) -> BreakableEntry {
         *self
     }
+
+    fn to_breakable_call_chain(self: Box<Self>) -> BreakableCallChainEntry {
```
fable-stripe: what do you think about doing option here, instead of unreachable

librubyfmt/src/render_targets.rs
```diff
@@ -150,43 +159,267 @@ impl AbstractTokenTarget for BreakableEntry {
...
     }
 
-    pub fn entry_formatting_context(&self) -> FormattingContext {
+    pub fn in_string_embexpr(&self) -> bool {
         self.context
+            .iter()
+            .any(|fc| fc == &FormattingContext::StringEmbexpr)
+    }
+
+    fn contains_hard_newline(&self) -> bool {
```
fable-stripe, line 194: does this need to iter any nested breakable entries?

librubyfmt/src/render_targets.rs
```diff
@@ -150,43 +159,267 @@ impl AbstractTokenTarget for BreakableEntry {
...
+pub struct BreakableCallChainEntry {
+    tokens: Vec<AbstractLineToken>,
+    line_numbers: HashSet<LineNumber>,
+    call_chain: Vec<CallChainElement>,
+    context: Vec<FormattingContext>,
+}
+
+impl AbstractTokenTarget for BreakableCallChainEntry {
+    fn to_breakable_entry(self: Box<Self>) -> BreakableEntry {
+        unreachable!()
```
fable-stripe: yeah I feel very strongly this should option, a programming mistake can cause us to reach an unreachable, which is what option unwrap is for

librubyfmt/src/format.rs
reese: I don't think we need to because `format_{do,brace}_block` should handle comment shifting internally (since they'll want to render the comments inside the block delimiters), but I'll add a comment here to make that explicit

librubyfmt/src/render_targets.rs
reese: I don't think so, at least not in the way it's currently used, because when we check this we also check whether `self.line_numbers` has multiple lines, which I think would handle having a multiline nested breakable. (FWIW this method is unchanged from before)

librubyfmt/src/format.rs
reese: Yeah this was kinda janky, I just combined it all into `next_args_list_must_use_parens` instead.

fixtures/small/single_line_method_call_chain_expected.rb
```diff
@@ -1,3 +1,15 @@
...
+This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain.build(
+  reason: Because::ThisCouldCauseUsSome::SeriousProblemsIfWeTriedToMultilineIt::EventThoughItActuallyDoesExtendBeyondTheLimit
+)
+
+This::Is::Some::SuperDuperLongConstantThatOnlyHasOneCallInTheChain
+  .build(
+    reason: Because::IPutThisOnMultipleLines
+  )
+
+this.surprisingly.can_break
```
fable-stripe, line 14: how does this work if this line is very long and there's a comment after it? can we add a sample

fables-tales (approved):
lgtm

PR #452, opened 2023-12-20
Represent binary operators as breakables

### LINKED ISSUE #379: Max line length isn't consistently respected
* Ruby version: 2.7.2
* Rubyfmt git sha: 4d090e0ea33cd6698a7eeb837a34c650e954a7a8

## Input file

```ruby
if Some::ReallyLongClassName::GoesHere::LetsMakeItEvenLonger.who_framed_roger_rabbit?(type: opts[:type], file_type: opts[:file_type])
end
```

## Rubyfmt's output

```ruby
if Some::ReallyLongClassName::GoesHere::LetsMakeItEvenLonger.who_framed_roger_rabbit?(type: opts[:type], file_type: opts[:file_type])
end
```

## This breaks because

This goes beyond the standard max line length of 120 characters (it's 133).

This is an inherent limitation in the way we've currently handled breakables, and I think we need to do a big ol' refactor to make this more consistent across the board, since it ends up breaking in several places to varying degrees. The short version is that in most cases, we check whether the breakable itself is longer than the line length limit, not whether it ends up passing the line limit _once it's been rendered_. This means in practice, we're not actually applying a _line_ length limit, but rather a _breakable_ length limit.

I haven't really dug into what the exact implementation for this would look like, but I think we'd end up doing something like this: during the render queue, track the line length of the current line and reset it at every hard newline, and then attempt to render breakables as single line and breaking if they go over. However, we'll have to be careful here -- we have to be sure that once we go to multiline, we still correctly track the line length _inside_ that breakable, because breakables nested inside of it may _also_ need to be further broken down. Take this as as an example:

```ruby
[ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, [ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing], ReallyLongThing, ReallyLongThing]
```

We've got two arrays, where the inner array is 119 characters long without including whitespace. _However_, since the outer array will render to multiline, the inner array will be indented two characters in, making the line 121 characters long, so it _should_ break to multiline, but it doesn't currently.

Current output:

```ruby
[
  ReallyLongThing,
  ReallyLongThing,
  ReallyLongThing,
  ReallyLongThing,
  [ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing, ReallyLongThing],
  ReallyLongThing,
  ReallyLongThing
]
```

Expected output:

```ruby
[
  ReallyLongThing,
  ReallyLongThing,
  ReallyLongThing,
  ReallyLongThing,
  [
    ReallyLongThing,
    ReallyLongThing,
    ReallyLongThing,
    ReallyLongThing,
    ReallyLongThing,
    ReallyLongThing,
    ReallyLongThing
  ],
  ReallyLongThing,
  ReallyLongThing
]
```

FWIW, this may also mean refactoring a bit of how we do some other "fake breakables"[0]  like method chains and binary operators, but just getting this to work for breakables would be a great start.

---

[0] it would be great to extend them to be _actual_ breakables, but blocks make that a bit difficult -- although not impossible

reese:
Closes #379 

This PR yanks out the hacked-together multilining strategy for binary operators and makes them breakables instead. This was largely easier than I though -- IIRC the original reason for not doing this was that call chains weren't also breakables, so that resulted in weird interactions, but since we did that in #443, this is a pretty straightforward change.

The way this works is that instead of our old hacked-together system, we can wrap the "top-level" binary in a breakable, and then for nested breakables, we just recurse into the formatting function instead of making a new breakable, which keeps the old behavior of keeping chained binary operations all at the same level.

This also fixed a number of accidental interactions that I wasn't aware of but that I think are actually bugs -- in the fixtures that we changed, this PR makes it so that binary operators now respect user-multiling in all cases, which they didn't used to do but which is consistent with the approach taken for other breakable items.

### DIFF
--- a/fixtures/small/binary_operators_expected.rb
+++ b/fixtures/small/binary_operators_expected.rb
@@ -1,6 +1,7 @@
 foo || bar
 
-foo || bar
+foo ||
+  bar
 
 if foo || bar
 end
@@ -27,7 +28,9 @@ a ||
   b &&
   c
 
-a || (b && c)
+a ||
+  (b &&
+    c)
 
 a ||
   b &&
--- a/fixtures/large/rspec_core_notifications_expected.rb
+++ b/fixtures/large/rspec_core_notifications_expected.rb
@@ -336,7 +336,9 @@ module RSpec::Core
           Formatters::Helpers.pluralize(failure_count, "failure")
         summary += ", #{pending_count} pending" if pending_count > 0
         if errors_outside_of_examples_count > 0
-          summary += (", " + Formatters::Helpers.pluralize(errors_outside_of_examples_count, "error") + " occurred outside of examples")
+          summary += (", " +
+            Formatters::Helpers.pluralize(errors_outside_of_examples_count, "error") +
+            " occurred outside of examples")
         end
 
         summary
--- a/fixtures/large/rspec_mocks_proxy_expected.rb
+++ b/fixtures/large/rspec_mocks_proxy_expected.rb
@@ -393,9 +393,8 @@ module RSpec
       # That's what this method (together with `original_unbound_method_handle_from_ancestor_for`)
       # does.
       def original_method_handle_for(message)
-        unbound_method = superclass_proxy && superclass_proxy.original_unbound_method_handle_from_ancestor_for(
-          message.to_sym
-        )
+        unbound_method = superclass_proxy &&
+          superclass_proxy.original_unbound_method_handle_from_ancestor_for(message.to_sym)
 
         return super unless unbound_method
         unbound_method.bind(object)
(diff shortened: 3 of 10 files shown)


fixtures/small/binary_operators_expected.rb
```diff
@@ -1,6 +1,7 @@
 foo || bar
 
-foo || bar
+foo ||
+  bar
```
reese, line 4: For basically all of these fixture updates, in the original versions, these were multilined, so keeping them multiline seems correct (and consistent with how we handle every other breakable)

fables-tales (approved):

PR #455, opened 2024-01-05
Prevent internal ruby failure on invalid braces

reese:
`""}` is invalid syntax, and while running it, Ripper tries to run `on_embexpr_end` to try and handle it as if it were the closing brace of an embexpr. However, in that case, we get a `NoMethodError` for trying to append to an array that isn't there (since we didn't call `on_embexpr_beg`. This is absolutely a weird edge case, but handling it means we correctly report this as a syntax error instead of an internal ruby error.

(There might be something to do here for a more sustainable solution, which is something like if we run into an internal ruby error, call the parser directly without `rubyfmt_lib` and see if it parses correctly, but I'll leave that as an idea for a later day.)

### DIFF
--- a/librubyfmt/rubyfmt_lib.rb
+++ b/librubyfmt/rubyfmt_lib.rb
@@ -477,7 +477,11 @@ class Parser < Ripper::SexpBuilderPP
 
   def on_embexpr_end(*args)
     # Append end line to make a StartEnd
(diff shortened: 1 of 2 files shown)


froydnj (approved):
@reese This seems like a good thing to have, but it'll need some rebasing after your recent test changes?

PR #458, opened 2024-01-09
Fix busted preview-release build by forcing full ruby builds

reese:
This PR fixes the `preview-release` build by removing the `ruby_checkout` caching. This was causing issues with the ruby build, but removing it fixes any apparent discrepancies in the build that was causing issues. This probably makes it a bit slower, but I don't think speed is of the utmost priority since it's only run post-merge anyways 🤷 

I verified this works by running it on this branch: [result](https://github.com/fables-tales/rubyfmt/actions/runs/7467397394/job/20320840420)

### DIFF
--- a/.github/workflows/preview-release.yaml
+++ b/.github/workflows/preview-release.yaml
@@ -6,7 +6,7 @@ on:
       - completed
   push:
     # Run only on trunk pushes that aren't a new tag release
-    branches: [trunk]
+    branches: [trunk, reese-preview-release]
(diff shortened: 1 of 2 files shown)


reese:
I forced a test build for the release and made sure it passed: https://github.com/fables-tales/rubyfmt/actions/runs/7670725309

fable-stripe (approved):

fables-tales (approved):

PR #463, opened 2024-02-21
Improve line length handling of end-of-line breakables in call chains

reese:
When calculating the line length for call chains, we do some hackery to make common expressions render nicely. Namely, we generally chop off the last do/end block in order to have e.g. `MyClass.do_stuff.each do ...` not split to multiple lines all the time. However, we were also chopping off the block parameters when we did this, which meant that instead of breaking the call chain when the params went over the line length, we'd break the block parameters, which led to a lot of situations of this shape:

```ruby
MyClass.do_stuff.pretend_this_is_longer.each do |
    my_arg
  |
  # body
end
```

That's pretty unfortunate -- instead, we should leave the block args intact whenever possible and consider them part of the call chain. We can still chop off the final block's body like we did before, which keeps most of the existing behavior, but this renders more nicely for most common cases as well.

### DIFF
--- a/librubyfmt/src/render_targets.rs
+++ b/librubyfmt/src/render_targets.rs
@@ -284,7 +284,13 @@ impl AbstractTokenTarget for BreakableCallChainEntry {
                         if *delims == BreakableDelims::for_block_params() {
                             // Wipe away the body of the block and leave only the params
                             *tokens = vec![tokens.first().unwrap().clone()];
+                        } else {
+                            // No params, so wipe away the whole thing
+                            *tokens = Vec::new();
                         }
+                    } else {
+                        // No params, so wipe away the whole thing
+                        *tokens = Vec::new();
                     }
                 }
             }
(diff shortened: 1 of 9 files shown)


librubyfmt/src/render_targets.rs
```diff
@@ -252,37 +252,66 @@ impl AbstractTokenTarget for BreakableCallChainEntry {
         // individual line and get _that_ max length.
         let mut tokens = self.tokens.clone();
         if tokens.len() > 2 {
-            if let Some(AbstractLineToken::ConcreteLineToken(ConcreteLineToken::End)) =
-                tokens.get(tokens.len() - 2)
-            {
-                // Pop off all tokens that make up the `do`/`end` block (but not `do`!),
+            let index = tokens.len() - 2;
```
reese, line 255: I wholly recognize that this method is getting to be quite convoluted (it was already a bit of a mess before, and this PR only adds to it), but call chains have a sufficient enough number of ~weird cases to warrant the complexity, and it's well tested enough that I feel like this is okay to leave for now. I'm not 100% sure the best way we'd clean it up without making some concessions on the final output anyways.

reese:
(For any Stripe reviewers, I'm bumping this since I believe this patch is in the rubyfmt version used in the Stripe codebase IIRC, so you'll probably want to land it at some point to not have weird drift when you want to pull in future improvements. I don't remember offhand why I had to bump the Stripe version with this patch, but 🤷 )

froydnj (approved):

PR #479, opened 2025-03-11
Remove `librubyfmt` caching from CI to fix ubuntu build

reese:
This caching has caused various headaches over the years (including currently breaking CI, but also #458 etc.) for reasons I never quite understood. It results in mysterious CI failures, and rebuilding Ruby doesn't actually take that long in CI anyways, so I think we should rip it out to make builds easier to repro in the case where they do fail.


froydnj (approved):
Yay!

PR #484, opened 2025-03-25
Set up`ruby-prism` crate behind a flag

reese:
I'd really like to start taking a crack at moving things over to Prism. However, I can't think of a particularly great "gradual cutover" scenario -- we've had to shove in so many one-off edge cases to support Ripper that the Prism integration will mostly end up being _very_ different. A lot of the core machinery will still be the same, but almost everything in `format.rs` will end up changing to a greater or lesser degree, since the way we hacked in line handling, heredocs, etc. is all going to be notably different (hopefully, much simpler!).

To that end, I didn't want to big-bang develop the Prism version, but instead have both Prism and the Ruby checkout exist side-by-side. This PR adds a `--prism` flag (as well as a `RUBYFMT_PRISM` environment variable) behind which we can hack on Prism support. This should make development much easier, and it will also probably be easier for folks to test this one their own codebases thoroughly before we rip out the Ruby checkout for good.

All this does so far is add the crate, a CLI flag, support for formatting integers/floats, and a simple smoke test to run in CI. As we add support for more and more nodes, we can add more tests to `test_prism.sh` and get complete test coverage.

Note that I didn't really add any nice user-facing errors to unsupported features in the Prism feature flag -- right now they're all just `todo!()` macros. I figured better messaging is not particularly helpful since we probably won't ever expose this in the documentation and likely will only ever need this for development purposes.


script/test_prism.sh
```diff
@@ -0,0 +1,21 @@
...
+# TODO(reese): uncomment these as we build out Prism support
+# ./script/tests/test_string_literals.sh
+# ./script/tests/test_array_literals.sh
+# ./script/tests/test_methods.sh
+# ./script/tests/test_cli_interface.sh
+# ./script/tests/test_c_main.sh
+# ./script/tests/test_error_handling.sh
+# ./script/tests/test_fixtures.sh
+FIXTURE_NAME=numbers ./script/tests/test_fixtures.sh
+# ./script/tests/test_formatting_locks.sh
```
froydnj, line 21: Long-term, I think we will be much better served by porting this into proper `#[test]` blocks, for parallelism and otherwise.  It's not obvious to me that we can do parallel tests in a single Rust process with things written as they are today, though.

froydnj (approved):
This seems like a reasonable way forward.

My idea for tackling this was ~porting Prism's (Ruby) translator into parser nodes into Rust, so that we could basically go from Prism AST -> the current internal representation that `rubyfmt` has after translating from Ripper.  But I think that leaves you in an awkward state once you're done, because that translation layer is awkward to rip out.  So this approach is better long-term, I think?

script/test_prism.sh
reese: Yeah I may do that after this, I'm not a huge fan of all the bash-based testing bits here, and I think things are going to get pretty awkward real quick trying to develop the two sides independently

PR #485, opened 2025-03-25
Bump to `v0.11.0`

reese:
It's been quite a while since we've made a non-prerelease release of `rubyfmt`, and since the last release in 2023, there are some non-trivial changes to the codebase and to the formatted output, so I think it's worth bumping us up a minor version. (Also, selfishly I'd love to have the current version of `rubyfmt` in `brew` so I don't have to keep downloading the release artifacts and shoving them into all of my personal projects.) Plus, if the next big chunk of work is updating this to Prism, that too will likely be another version bump, so I'd like to differentiate all of those as separate versions.

After this goes in, I think I have the permissions to make a release in the Github UI, which I _think_ will make a tag that will trigger the [release workflows](https://github.com/fables-tales/rubyfmt/blob/reese-v-11/.github/workflows/release.yaml). We haven't had a chance to really try those out I don't think, so this may be a bit of trial-and-error, but now's as good a time as any!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -486,7 +486,7 @@ dependencies = [
 
 [[package]]
 name = "rubyfmt"
-version = "0.10.0-pre"
+version = "0.11.0-pre"
 dependencies = [
  "backtrace",
  "cc",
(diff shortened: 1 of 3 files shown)


froydnj (approved):
`:letsgooo:`

PR #488, opened 2025-03-25
Migrate test fixtures to `cargo test`

reese:
This moves our "actual vs expected" test fixtures from running in bash to actual `#[test]` integration tests. I made a quick macro for this and then explicitly wrote all of them out -- we could in theory have a generated test file for this if we'd rather, but I kinda liked the explicitness of just having them be written out.

After this, I'd like to move all the other tests also into Rust instead of bash, but since these are by far the largest in number, I figured it'd be good to do them as their own PR first.


tests/fixtures_test.rs
```diff
@@ -0,0 +1,579 @@
...
+
+fixture!(
+    test_rspec_core_notifications,
+    "large/rspec_core_notifications"
+);
+
+// TODO: This test fails in debug but works in release
+// due to some newline shenanigans, we should investigate why
+#[cfg(not(debug_assertions))]
+fixture!(test_rspec_mocks_proxy, "large/rspec_mocks_proxy");
```
reese, line 579: Interesting discovery that these tests were failing some debug assertions with a plain `cargo test`, but `cargo test` doesn't have that assertion and they pass, so I've just disabled them in debug builds for now. CI runs `cargo test --release`, so they're still run eventually.

froydnj (approved):
❤️

PR #490, opened 2025-03-27
Convert formatting locks tests into plain ol' fixtures

reese:
Formatting locks are essentially real-world codebases that were used as tests. The script that tested them did a `git clone` of the repo, checked out a given commit, and then formatted select files and compared the formatted output to an md5 sha as a "lock."

Personally, I'd rather just have these files exist as fixtures. Whenever the lock changes, all you get is a failure saying the md5 changed -- if you want details on _what_ changed (to make sure your changes are right), you probably would just have to go check out the repo yourself. Sure, this adds a few extra thousand lines to the codebase, but they were essentially already there anyways, now it's just more explicit!

With fixtures, you (1) can see the visual diff in a PR that updates the locks, (2) you don't have to `git clone` and `git checkout` in a test, and (3) we don't have to rewrite all of the git clone procedures into Rust if we really want this to be usable in `cargo test` -- we can use the existing test infra I already wrote.


froydnj (approved):
Can't wait until `rubyfmt` embeds shopify-core for testing purposes.

reese:
_In my next PR: `pay_server_stress_test.rs`_

PR #497, opened 2025-03-31
⛓️  Add basic support for call chains with Prism

reese:
This adds support for call chains, as well as various other node types needed to get some important fixtures (like the methods stress test and `fixtures/small/single_line_method_call_chain_actual.rb`). I also did some manual checking in some other fixture files, but those depend on other complex nodes (strings, blocks mostly) that I wanted to split into their own PRs.

The primary difference between the Ripper version and this one is that this one "precalculates" user-multilining instead of calculating it when rendering the breakable. This is primarily to avoid having to hold on to copies of the Prism nodes into the render queue. Maintaining both versions is done with an intermediate struct that adds a little bit of clunkiness for now, but it makes it easy to be able to rip out once the Ripper implementation.

### DIFF
--- a/librubyfmt/src/file_comments.rs
+++ b/librubyfmt/src/file_comments.rs
@@ -1,10 +1,10 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::mem;
 
 use crate::comment_block::CommentBlock;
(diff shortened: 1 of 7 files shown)


froydnj (approved):
I skimmed the Rust code, but I'm mostly assuming that "passes tests == good" here.

I think you will have to reconcile this with #512.

PR #498, opened 2025-03-31
Convert CLI tests to Rust

reese:
On the tin -- one step closer to having our tests just be a simple `cargo test` invocation!

Third in the series of #488 and #491


froydnj (approved):
Probably worth writing a test helper or two for a lot of the `Command` boilerplate at some point.

Issue #454, opened 2023-12-20
Characters in string becomes Unicode literal notation


ryym:
* Ruby version: `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]`
* Rubyfmt git sha: e00d2ab (v0.10.0)

## Input file

```ruby
# ok
# hello 😀

# ok
'hello'
'😀'

# NOT OK
'hello 😀'
```

## Rubyfmt's output

```
# ok
# hello 😀

# ok
"hello"
"😀"

# NOT OK
"hello \u{1F600}"
```

## This breaks because

It seems that Rubyfmt outputs non-ASCII characters in a string as Unicode literal notation when there are ASCII characters in the same string. The problem occurs for non-emoji characters too, such as `あいうえお` , `漢字` , `유니코드` , etc.

reese:
The newest version of `rubyfmt` is now much less aggressive about these escapes, so this should now be fixed in `v0.12.0`:

```bash
% cat hi.rb      
# ok
# hello 😀

# ok
'hello'
'😀'

# NOT OK
'hello 😀'

'漢字 also work here'                                                                                                                   

% cargo run --release hi.rb
# ok
# hello 😀

# ok
"hello"
"😀"

# NOT OK
"hello 😀"

"漢字 also work here"
```

PR #501, opened 2025-04-02
🧶 Support string literals with Prism

reese:
This PR gets us far enough to pass the string literals stress test with Prism, meaning that the Prism implementation can generally correctly handle most heredocs, single quote strings, double quote strings, and percent-literal strings. This implementation probably isn't 100% perfect, and there's likely some additional future testing that will need to be done to fully test all of this, but this represents a solid first pass at the core functionality.

This implementation attempts to be faithful to the way that strings were handled in `rubyfmt_lib.rb`, but that implementation relies heavily on `eval` and `String#inspect`, neither of which we'll have access to. This implementation essentially re-implements parts of `String#inspect`, but it's not going to be 100%, nor should it be -- `inspect` causes all sorts of (IMO) unexpected results in issues like #454 where things like English characters are mixed with emoji/kanji/etc., but this is essentially a non-issue in the prism implementation. That said, once this is more fully featured, there may be edge cases not in the fixtures, but I _think_ this will cover most cases even in its current form.

### DIFF
--- a/librubyfmt/Cargo.toml
+++ b/librubyfmt/Cargo.toml
@@ -10,6 +10,7 @@ build = "build.rs"
 [dependencies]
 serde = { version = "1.0", features=["derive"] }
 libc = "0.2.68"
+fancy-regex = "0.14.0"
 ripper_deserialize = { path = "ripper_deserialize" }
 lazy_static = "1.4.0"
 log = { version = "0.4.8", features = ["max_level_debug", "release_max_level_warn"] }
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -113,7 +113,9 @@ pub fn format_node(ps: &mut dyn ConcreteParserState, node: prism::Node) {
         Node::InstanceVariableOrWriteNode { .. } => todo!(),
         Node::InstanceVariableReadNode { .. } => todo!(),
         Node::InstanceVariableTargetNode { .. } => todo!(),
-        Node::InstanceVariableWriteNode { .. } => todo!(),
+        Node::InstanceVariableWriteNode { .. } => {
+            format_instance_variable_write_node(ps, node.as_instance_variable_write_node().unwrap())
+        }
         Node::IntegerNode { .. } => format_integer_node(ps, node.as_integer_node().unwrap()),
         Node::InterpolatedMatchLastLineNode { .. } => todo!(),
         Node::InterpolatedRegularExpressionNode { .. } => todo!(),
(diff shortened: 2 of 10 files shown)


librubyfmt/Cargo.toml
```diff
@@ -12,6 +12,7 @@ serde = { version = "1.0", features=["derive"] }
 serde_json = "1.0.40"
 backtrace = "0.3.45"
 libc = "0.2.68"
+fancy-regex = "0.14.0"
```
reese, line 13: N.B. the `regex` crate doesn't support lookbacks, which we originally used in the `rubyfmt_lib` implementation, hence this uses `fancy-regex` instead (which is really just the `regex` crate plus lookbacks and other regex features that operate in greater-than-linear time).

librubyfmt/src/format_prism.rs
```diff
@@ -235,6 +242,184 @@ fn format_statements(ps: &mut dyn ConcreteParserState, statements_node: prism::S
     );
 }
 
+fn format_string_node(ps: &mut dyn ConcreteParserState, string_node: prism::StringNode) {
+    if ps.at_start_of_line() {
+        ps.emit_indent();
+    }
```
froydnj: I thought we were doing all of these at the start of `format_node` nowadays, so we can delete this bit and the matching newline code in this function and elsewhere?

librubyfmt/src/format_prism.rs
```diff
@@ -235,6 +242,184 @@ fn format_statements(ps: &mut dyn ConcreteParserState, statements_node: prism::S
...
+    ps: &mut dyn ConcreteParserState,
+    interpolated_string_node: prism::InterpolatedStringNode,
+) {
+    if ps.at_start_of_line() {
+        ps.emit_indent();
+    }
+
+    let is_heredoc = interpolated_string_node
+        .opening_loc()
+        .map(|s| loc_to_string(s).starts_with("<"))
```
froydnj, line 317: `loc_to_str` at some point might be nice, also for the `loc_to_string.trim.to_string` patterns in this file.

librubyfmt/src/format_prism.rs
```diff
@@ -235,6 +242,184 @@ fn format_statements(ps: &mut dyn ConcreteParserState, statements_node: prism::S
...
+    //   "bar"
+    // ```
+    // as an interpolated node with the contents `"foobar"` (in two `parts` of "foo" and "bar").
+    // To detect this, we can look for any `InterpolatedStringNode` that has multiple `parts`
+    // and isn't a heredoc.
+    let is_backslash_string_interpolation = !is_heredoc
+        && interpolated_string_node
+            .parts()
+            .iter()
+            .all(|node| node.as_string_node().unwrap().opening_loc().is_some());
```
froydnj: This `unwrap()` does not feel guaranteed to be safe, but I don't have an obvious counterexample available.  Something like `.map(|s| s.opening_loc).is_some()` would be safer?

froydnj (approved):
Approving this to unblock, but I think you do want to fix the `at_start_of_line` checks before merging.

PR #502, opened 2025-04-02
Support basic method defs with Prism

reese:
On the tin -- this adds support for most method definitions with Prism, including the various argument types (required, optional, keyword, block, etc.).

This also makes the Prism tests an explicit test file, instead of having them be in a separate `test_prism.sh` script. Now that we have all of our test in Rust instead of as bash files, having them in their own file will be easier than trying to craft some crazy `cargo test` invocation. It's currently just a duplicate of the fixtures test file with all the unsupported stuff commented out, which will serve as a burn-down list of fixtures to work through that we can uncomment as we get to them.


.github/workflows/ci.yml
```diff
@@ -46,32 +46,6 @@ jobs:
           ./script/lints/lint_fixtures.sh
           ./script/lints/lint_scripts.sh
 
-  prism:
```
fables-tales, line 49: ?

froydnj (approved):

librubyfmt/src/format_prism.rs
```diff
@@ -1,14 +1,20 @@
...
-use crate::{parser_state::ConcreteParserState, util::loc_to_string};
+use crate::{
+    delimiters::BreakableDelims,
+    format::SpecialCase,
+    parser_state::{ConcreteParserState, FormattingContext, RenderFunc},
+    types::SourceOffset,
+    util::{const_to_string, loc_to_string},
+};
 
 pub fn format_node(ps: &mut dyn ConcreteParserState, node: Node) {
```
fables-tales, line 11: I think this is from `ruby_prism`, but we imported `*` so it's hard to tell. Can we change the `*` import to either qualifying each import, or spamming `ruby_prism` prefixes everywhere?

fables-tales: to be clear: this is a fable preference, I hate `*` imports

.github/workflows/ci.yml
reese: At first I was just gonna have a prism-specific job to run tests, but instead I folded it all into actual explicit tests, so we don't need the separate job anymore

librubyfmt/src/format_prism.rs
reese: Yeah I can change them to explicit imports -- I figure we're going to import ~everything from `ruby_prism` anyways, but I'm fine to have them all be explicit imports (unless we really want `ruby_prism::` on everything, which seemed like it was just gonna make confusing method signatures, but :shrug:)

.github/workflows/ci.yml
reese: That is to say: the prism fixtures are run in `tests/prism_fixtures_test.rs` now using the `--prism` flag instead of some spooky environment variable shenanigans

librubyfmt/src/format_prism.rs
fables-tales: could alias import it, `use rust_prism as rp` ?

.github/workflows/ci.yml
fables-tales: cool sg

librubyfmt/src/format_prism.rs
reese: #507

PR #506, opened 2025-04-02
Bump Ubuntu GH action runners 20.04->22.04

reese:
The Ubuntu 20.04 runners are going to be deprecated in the next few weeks (https://github.com/actions/runner-images/issues/11101), so we should move off of them.

cc @froydnj I recall us using 20.04 specifically because we needed that to be built for Stripe. Not sure what the current state of the world is there, but a heads up just in case the specifics here affect y'all's ability to use the release artifacts. Currently GH runners will only support 22.04 and 24.04, and our other jobs also run `ubuntu-latest` (24.04), so I just picked 22.04 to have coverage on both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,11 +80,11 @@ jobs:
         include:
           - os: ubuntu-latest
             target: native
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             target: native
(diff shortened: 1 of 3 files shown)


froydnj (approved):
_grumbles about deprecation of old platforms for CI_

I think 22.04 should work out OK for us.

PR #508, opened 2025-04-03
Update some dependencies to consolidate multiple crate version deps

reese:
A few of our dependencies require us to have deps on the same crate at multiple versions, meaning we have to compile both versions etc. This PR bumps a few of our crates to newer versions (`syn` being the only major version, the others are just patch bumps) to consolidate those.

(And relatedly: `cargo tree --duplicate` is cool)


froydnj (approved):
🧹

PR #511, opened 2025-04-04
Fix const_path_ref class definitions

reese:
I mistakenly used `name()` for the class name earlier, when in actuality we should be using `constant_path()` to get the parent classes as well in the case of a const path.


froydnj (approved):
Doh!

PR #512, opened 2025-04-04
Handle indents/newlines in `format_node` instead of hand-writing it in every node impl

reese:
This is an attempt to rectify something soon rather than later that was inherited from the Ripper implementation. In the original Ripper implementation, a ton of node types have `if ps.at_start_of_line() { ps.emit_indent() }` and `if ps.at_start_of_line() { ps.emit_newline() }` at the beginning and end of _so_ many nodes, and the same is beginning to be true for Prism. This is pretty much always unnecessary as long as we declare `ps.with_start_of_line` in the appropriate places, which we should do!

The prism tree is constructed such that I think each node should have enough information to handle it's line-start appropriately, so we'll save ourselves a ton of boilerplate by putting this at the beginning of `format_node` and letting that be the source of truth.

(I'd also recommend hiding whitespace changes in this diff, that'll make it a bit clearer)

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -11,6 +11,14 @@ use crate::{
 
 pub fn format_node(ps: &mut dyn ConcreteParserState, node: prism::Node) {
     use prism::Node;
+    // StatementsNode is the only real "wrapper" node, meaning it purely contains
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Makes sense to me!

PR #514, opened 2025-04-05
Use github hosted ARM runners and remove cross-compilation infra

reese:
A while back when we needed ARM builds at Stripe, there wasn't yet a Github-hosted ARM runner, so #436 added bits to cross-compile from x86 Linux. Last year, GH created an [ARM Linux runner](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md) that is [now available to the public](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), so we can just use that!

The main benefit for us, aside from tooling simplicity, is that our previous cross-compilation infra didn't actually run tests on arm, so we were sort of just assuming that the cross-compiled binaries worked correctly (which, thankfully, they always did!). This let's us actually run tests on ARM, which is an extra layer of security.

This PR removes all the cross-compilation bits and instead replaces them with a plain ol' `ubuntu-22.04-arm` runner. I can't really test the preview-release and release actions, but all that changed is the artifact names, which I think will be easy enough to fix if things go south.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,13 +53,9 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            target: native
           - os: ubuntu-22.04
-            target: native
           - os: macos-latest
-            target: native
-          - os: ubuntu-22.04
-            target: aarch64-unknown-linux-gnu
+          - os: ubuntu-22.04-arm
     steps:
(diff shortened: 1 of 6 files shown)


froydnj (approved):
Let's try it and see how it goes!  Very worth deleting that gross build code.

PR #516, opened 2025-04-06
🔪  Remove old testing bash scripts

reese:
We now, at long last, have our test suite able to be run just as `cargo test [--release]`! To that end, we no longer really need all of these bash scripts, so let's 🔪 them


CONTRIBUTING.md
```diff
@@ -7,9 +7,9 @@
 
 ## Running the tests
 
-`./script/test.sh`
+`cargo test`
 
-in particular `./script/tests/test_fixtures.sh` is what I use for ongoing
+in particular `cargo test --test fixtures_test` is what I use for ongoing
```
reese, line 12: (This whole file probably needs some brushing up at some point, but that's for another day)

froydnj (approved):

PR #523, opened 2025-04-09
Format super nodes with Prism

reese:
What it says -- this adds support for `super` nodes, which in Prism are two separate node types: `ForwardingSuperNode` which is `super` without parens, and `SuperNode` which is `super()` with parens (and possibly arguments etc.)

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -737,8 +737,7 @@ fn format_class_node(ps: &mut dyn ConcreteParserState, class_node: prism::ClassN
     );
 
     if let Some(superclass) = class_node.superclass() {
-        ps.emit_ident("<".to_string());
-        ps.emit_space();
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -737,8 +737,7 @@ fn format_class_node(ps: &mut dyn ConcreteParserState, class_node: prism::ClassN
     );
 
     if let Some(superclass) = class_node.superclass() {
-        ps.emit_ident("<".to_string());
-        ps.emit_space();
+        ps.emit_ident(" < ".to_string());
```
reese, line 740: This was just a silly thing from before I had implemented enough stuff for the tests with superclasses -- this is just adding the missing space before the `<`

froydnj: Doh.

froydnj (approved):

PR #525, opened 2025-04-09
Fix comments in empty blocks in Prism impl

reese:
Blocks whose contents are only comments run into all sorts of weird edge cases, particularly with respect to [brace blocks](https://github.com/fables-tales/rubyfmt/blob/373338d1c356d7906c16a9d136549b36b24599e5/librubyfmt/src/format.rs#L2964-L2970). This PR handles those edge cases better, at least well enough to pass the fixtures we put specifically for those edge cases


froydnj (approved):
This PR has a very low lines-changed-to-fixtures-working ratio.  I demand better!

😉 ❤️

PR #526, opened 2025-04-09
Fix spacing for `LocalVariableWriteNode`s in Prism

reese:
~To honor froydnj's wishes for a higher lines-changed-to-fixtures-passing ratio, I present this PR as penance after #525 -- 5 fixtures for 1 line!~

Doh.


froydnj (approved):
This is better!  `:more:`

PR #534, opened 2025-04-13
[prism] Implement Range operators

reese:
What it says -- this implements `..` and `...` operators, plus it fixes a small issue with paren nodes being incorrectly multilined all the time.


librubyfmt/src/format_prism.rs
```diff
@@ -1115,15 +1124,18 @@ fn format_call_node(
...
-                                    if call_operator != *"::" {
-                                        ps.emit_collapsing_newline();
-                                        ps.emit_soft_indent();
+
+                                    // `call_operator_loc` is the `.`/`::`/`&.` etc.
+                                    // it may be None in the case of arefs, e.g. foo[bar]
+                                    let call_operator =
+                                        element.call_operator_loc().map(|loc| loc_to_string(loc));
+                                    if let Some(call_operator) = call_operator {
+                                        if call_operator != *"::" {
```
froydnj, line 1133: 😭

froydnj (approved):

librubyfmt/src/format_prism.rs
reese: > reese: there must be an operator

_Narrator: there was not an operator_

PR #539, opened 2025-04-14
[ci] Add image "fingerprint" to Cargo cache key

reese:
One of our recent trunk builds failed because of some mysterious glibc version mismatch. Once I cleared the action cache, it started passing again, and while I couldn't track down a specific runner change that caused this, that implies to me that the underlying glibc might have been updated for the Ubuntu arm runner, but the cached version used the old version and failed when trying to link to a newer version.

To future-proof this, this PR adds a "fingerprint" to the cache key. The methodology for this is a little bit convoluted, if only because I couldn't find anything like an image sha or commit ID that's publicly available for runners, so what I did instead was to make a "fingerprint" of the OS version and a sha of all that images installed packages (with `dpkg` for linux and `brew` for mac), which hopefully should catch things like glibc changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,12 +35,20 @@ jobs:
           toolchain: 1.85.0
           components: clippy
           override: true
+      - name: Generate system fingerprint for cache key
+        id: system-fingerprint
+        run: |
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+            echo "SYS_FINGERPRINT=$(dpkg -l | md5sum | cut -d ' ' -f1)" >> $GITHUB_ENV
+          else
+            echo "SYS_FINGERPRINT=$(sw_vers | md5 | cut -d ' ' -f1)_$(brew list --versions | md5 | cut -d ' ' -f1)" >> $GITHUB_ENV
+          fi
       - uses: actions/cache@v4
         with:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ runner.os }}-cargo-2-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-2-${{ hashFiles('**/Cargo.lock') }}-${{ env.SYS_FINGERPRINT }}
       - run: cargo clippy --all --all-features -- -D warnings
 
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -35,12 +35,20 @@ jobs:
           toolchain: 1.85.0
           components: clippy
           override: true
+      - name: Generate system fingerprint for cache key
+        id: system-fingerprint
+        run: |
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+            echo "SYS_FINGERPRINT=$(dpkg -l | md5sum | cut -d ' ' -f1)" >> $GITHUB_ENV
```
froydnj, line 42: Wow, that's....aggressive.

froydnj (approved):
We can't use the distro codename or something like that?  Depending on all the packages installed on the system is a big hammer -- I can believe that package versions spanning OS versions might have problems (though glibc is pretty robust...?) but package versions within the same OS version ought to be pretty compatible.

I guess I could believe that something got compiled against a glibc on Noble and reusing those objects on Focal might not go so well...

I'm sure there's somebody out there reading this that's thinking we should just be using Nix instead.

.github/workflows/ci.yml
reese: Yeah this is almost certainly overly aggressive, but it's not exactly clear to me why this failed in the first place and spurious trunk builds make me sad so I went with a bigger hammer over a smaller one.

It's entirely possible it was like a noble/focal mismatch, and we could instead start with just tacking on the distro codename or number, although it's not entirely clear to me why that wouldn't have failed earlier or why it only failed for the 22.04 ubuntu runner and not 24.04 (since I assume that failure mode would cut both ways?).

Or we could just switch to Nix.

reese: Also if it's potentially _just_ a glibc version thing, we could also not be so aggressive about hashing literally every dependency and just care about glibc -- I doubt we'll end up with that many system dependencies in the future outside of that.

reese: What's surprising is that the build before the failures, the failed builds, and the builds afterwards all do you use the same image version (`20250406.1.0`), so it's actually not even clear to me if this would actually have fixed them, since I'm assuming all of those images would have the same glibc etc. `:thonking_sad:`

reese: So with that said, I think I may go the route of "let's actually just wait and see" and close this out for now, since I'm not sure this actually fixes anything and might just be more chunder/complexity. If things start mysteriously failing again, this might get revived in some other form, but it might have just been some one-off funkiness -- hard to know since we don't have full control over the build images 🤷

PR #540, opened 2025-04-14
Migrate to Rust edition 2024

reese:
`rubyfmt` has always been on the 2018 edition, but getting on to the 2024 edition is mostly just a matter of adding some `unsafe` blocks (which I just accepted from the autocorrect) and then reformatting.

The impetus for this is mostly just that it fixes some issues in `rustfmt` that have always annoyed me (like the chained tuple formatting as `foo.0 .1` with that awkward space), plus it'll make it easier for us to adopt features in future rust versions as they get stabilized.


froydnj (approved):
`:whatyearisit:`

PR #541, opened 2025-04-14
[prism] Support method modifier calls

reese:
This adds special handling for method modifiers like `private`/`protected`/etc., which are special-cased to always format like `private def foo` instead of using parens.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1097,22 +1097,45 @@ fn format_call_node(
             );
         }
         if let Some(arguments) = call_node.arguments() {
-            let delims = if is_aref {
-                BreakableDelims::for_array()
+            // For callers where the only arg is a def node,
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1097,22 +1097,45 @@ fn format_call_node(
...
+            {
+                ps.emit_space();
+
+                let def_node = arguments
+                    .arguments()
+                    .iter()
+                    .next()
+                    .unwrap()
+                    .as_def_node()
+                    .unwrap();
```
froydnj, line 1119: Faster `let_chains` stabilization, please: https://github.com/rust-lang/rust/issues/53667

froydnj (approved):

librubyfmt/src/format_prism.rs
reese: Literally the exact reason I made #540 😆

PR #542, opened 2025-04-15
[prism] Fix some issues with heredoc rendering

reese:
(Editorial note: Man I forgot how annoying heredocs are.)

This PR fixes various issues with rendering heredocs. The primary changes are that it fixes indentation issues with squiggly heredocs (since they are ✨special✨  and their indentation is relative to both the current indentation level _and_ the indentation of the least-indented line), plus it fixes the way they get rendered so that they work correctly inside argument lists. The latter is mostly fixed by just using all the `render_heredocs` machinery that already existed -- I just didn't implement it that way the first time around.

### DIFF
--- a/librubyfmt/src/format.rs
+++ b/librubyfmt/src/format.rs
@@ -1544,7 +1544,7 @@ pub enum StringType {
     Regexp,
 }
 
-pub fn format_inner_string(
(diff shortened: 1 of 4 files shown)


froydnj (approved):
`:so-it-begins:`

PR #550, opened 2025-04-21
[prism] Implement hash node

reese:
What it says -- this is pretty much just a direct port of `format_hash` from the Ripper implementation, they should be practically equivalent in Prism.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -2085,8 +2085,50 @@ fn format_global_variable_write_node(
     todo!()
 }
 
-fn format_hash_node(_ps: &mut dyn ConcreteParserState, _hash_node: prism::HashNode) {
-    todo!()
+fn format_hash_node(ps: &mut dyn ConcreteParserState, hash_node: prism::HashNode) {
+    ps.with_start_of_line(
+        false,
+        Box::new(|ps| {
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2085,8 +2085,50 @@ fn format_global_variable_write_node(
...
+                ps.breakable_of(
+                    BreakableDelims::for_hash(),
+                    Box::new(|ps| {
+                        ps.emit_soft_indent();
+                        format_list_like_thing(
+                            ps,
+                            hash_node.elements(),
+                            hash_node.closing_loc().end_offset(),
+                            false,
+                        );
```
froydnj, line 2125: This eventually needs to replicate the logic in `format_assocs` for hashrocket syntax etc., right?

froydnj (approved):

librubyfmt/src/format_prism.rs
reese: Yeah, it'll probably end up having something like [this](https://github.com/fables-tales/rubyfmt/blame/0d1d1e97e4f596d553d6bb7722393dcfce53c979/librubyfmt/src/format_prism.rs#L1822-L1832), but there's some other things to fix with symbols etc. that I need to do to get all of those related fixtures passing

PR #551, opened 2025-04-21
[prism] Fix endless methods spacing

reese:
There are some cases in Prism where a `StatementsNode` is used, but there can only be a single statement that requires separate treatment. Endless methods are one such example, where that single treatment is treated more like the right side of an assignment instead of like a list of statements (e.g. a regular method body, a class def, etc.),  so this PR yanks that single method out of the StatementsNode and renders it directly to fix some weird spacing issues.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1095,7 +1095,15 @@ fn format_def_body(ps: &mut dyn ConcreteParserState, def_node: prism::DefNode) {
                         false,
                         Box::new(|ps| {
                             if let Some(body) = def_node.body() {
-                                format_node(ps, body);
+                                format_node(
+                                    ps,
+                                    body.as_statements_node()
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1095,7 +1095,15 @@ fn format_def_body(ps: &mut dyn ConcreteParserState, def_node: prism::DefNode) {
...
                             if let Some(body) = def_node.body() {
-                                format_node(ps, body);
+                                format_node(
+                                    ps,
+                                    body.as_statements_node()
+                                        .expect("Endless methods must have a body, and method definitions are always a Statements node")
+                                        .body()
+                                        .iter()
+                                        .next()
+                                        .expect("Endless methods must have exactly one expression in their body")
```
froydnj: Nit: this would be more resilient to catching future prism changes if we could also assert `None`-ness about `.next()`'ing the iterator again.

froydnj (approved):

PR #554, opened 2025-04-25
[prism] Support `next` keyword

reese:
What it says -- this also slightly modifies some of the fixtures, since understandably Prism considers `yield` outside of definitions and `next` outside of blocks to be invalid syntax, so this just wraps the old fixtures in blocks/defs as needed.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1637,6 +1637,13 @@ fn format_block_parameters_node(
     ps: &mut dyn ConcreteParserState,
     block_parameters_node: prism::BlockParametersNode,
 ) {
+    // Exit early if there's no params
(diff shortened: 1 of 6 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1637,6 +1637,13 @@ fn format_block_parameters_node(
     ps: &mut dyn ConcreteParserState,
     block_parameters_node: prism::BlockParametersNode,
 ) {
+    // Exit early if there's no params
+    if node_list_is_empty(&block_parameters_node.locals())
+        && block_parameters_node.parameters().is_none()
+    {
+        return;
+    }
+
```
reese, line 1646: This is just to match behavior of the Ripper implementation, which doesn't render empty block params (`||`)

froydnj (approved):

PR #555, opened 2025-04-25
[prism] Support block args (`&blk`)

reese:
So block args are a bit complicated, but that's because Prism represents `foo(&blk)` and `foo { || }` basically the same way. In the former case, they're not included in the `CallNode.arguments()` list, as one may expect, but rather only in the `block()` node, so you have to sort of backport them into the arguments list if there's an argument list, treat them as the sole argument if there's _not_ an args list, and otherwise treat regular brace or do-end blocks normally if they're present.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1347,8 +1347,7 @@ fn format_call_node(
                                 // we want it to be a part of the comma-separated list
                                 if let Some(block_argument_node) = call_node
                                     .block()
-                                    .map(|block_node| block_node.as_block_argument_node())
-                                    .flatten()
+                                    .and_then(|block_node| block_node.as_block_argument_node())
                                 {
                                     if has_arguments {
                                         ps.emit_comma();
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1331,21 +1339,50 @@ fn format_call_node(
...
+                );
+            // If there's an arguments node, we've handled this block arg with
+            // the rest of the args (since it's included in the comma-separated
+            // args list), otherwise the only argument is the &blk node, so we
+            // have to handle that here separately
+            } else if call_node.arguments().is_none() && block.as_block_argument_node().is_some() {
+                ps.breakable_of(
+                    BreakableDelims::for_method_call(),
+                    Box::new(|ps| {
+                        format_block_argument_node(ps, block.as_block_argument_node().unwrap());
```
reese, line 1381: I am once again asking Rust to stabilize if-let chains

froydnj (approved):

PR #561, opened 2025-10-22
[prism] add formatting for class and instance variables

froydnj:
There is not great test coverage here, but ideally the cargo-culting from local variables will be sufficient.  I did try to add some very basic tests; I'm sure that some of the more complex tests will eventually exercise this stuff.

### DIFF
--- a/fixtures/small/ivar_actual.rb
+++ b/fixtures/small/ivar_actual.rb
@@ -1,0 +1,3 @@
+class Ivar
+  @converter = NullConverter
+end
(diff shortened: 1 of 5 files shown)


elliottt (approved):
Looks good! What do you think about adding some more test cases that differ in the actual/expected structure?

froydnj:
> Looks good! What do you think about adding some more test cases that differ in the actual/expected structure?

I will try to do some of this in a separate PR; I think the issue is more widespread than just these few tests.

PR #563, opened 2025-11-01
rewrite `gather_heredocs` in a functional manner

froydnj:
This is just a little bit more concise, and probably a bit more efficient to boot.

r? @elliottt

### DIFF
--- a/librubyfmt/src/parser_state.rs
+++ b/librubyfmt/src/parser_state.rs
@@ -902,11 +902,7 @@ impl BaseParserState {
         if self.heredoc_strings.is_empty() {
             None
         } else {
-            let mut hds = vec![];
-            while let Some(element) = self.heredoc_strings.pop() {
-                hds.push(element);
-            }
-            Some(hds)
+            Some(self.heredoc_strings.drain(..).rev().collect())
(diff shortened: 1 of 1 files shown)


librubyfmt/src/parser_state.rs
```diff
@@ -902,11 +902,7 @@ impl BaseParserState {
         if self.heredoc_strings.is_empty() {
             None
         } else {
-            let mut hds = vec![];
-            while let Some(element) = self.heredoc_strings.pop() {
-                hds.push(element);
-            }
-            Some(hds)
+            Some(self.heredoc_strings.drain(..).rev().collect())
```
elliottt, line 905: Could also do `std::mem::take` of the `heredoc_strings` and then reverse them, but perhaps it's good to leave the allocation around?

elliottt (approved):

librubyfmt/src/parser_state.rs
froydnj: I think it's probably a good thing that we leave the memory around for later, but I like the `std::mem::take` suggestion and will keep it in mind for next time!

PR #565, opened 2025-11-01
collapse `ConcreteParserState` and `BaseParserState` into `ParserState`

froydnj:
`ParserState` was split into `ConcreteParserState` and `BaseParserState` in 05807c966612ebb2512c617381229d24cc16cd0a , which required the introduction of boxing closures passed to `ConcreteParserState` methods.  (You can't just use `impl FnOnce(&mut dyn ConcreteParserState)` or similar, or `rustc` yells at you.)  I don't think the split has been successful:

1. `ConcreteParserState` is pretty big.
2. We have trait methods that explicitly know about the presence of `BaseParserState` (to be fair, this is probably a bit of oversight, and those methods could just as easily be modified to use `ConcreteParserState`).
3. The boxing of closures is verbose and inefficient.
4. The naming is a little strange -- I'm sure the names were chosen to make the refactoring easier, but it seems odd to me to have the trait named something `Concrete*`.
5. I don't think we ever intend to have another implementation of `ConcreteParserState`.
6. Probably some other things that I am overlooking.

To be fair, I do think `ConcreteParserState` did provide a reasonable division between the "API" of the parser state and the strictly private helper functions on `BaseParserState`.  But I think that can be accomplished mostly as easily with visibility attributes, and doesn't require duplication of e.g. method signatures for no particular reason.

This PR undoes the naming part of the above; I have follow on patches to undo the closure boxing.  I admit to being lazy in just `pub (crate)`'ing all the functions on `BaseParserState`, and there are probably some that could be deleted.

wdyt @elliottt ?

### DIFF
--- a/librubyfmt/src/format.rs
+++ b/librubyfmt/src/format.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 
 use crate::delimiters::BreakableDelims;
 use crate::heredoc_string::HeredocKind;
-use crate::parser_state::{ParserState, FormattingContext, RenderFunc};
+use crate::parser_state::{FormattingContext, ParserState, RenderFunc};
 use crate::render_targets::MultilineHandling;
 use crate::ripper_tree_types::*;
 use crate::types::LineNumber;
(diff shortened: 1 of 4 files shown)


fables-tales:
the original intent here was that the notion of concrete tokens vs abstract tokens and there was going to be some like cases where we didn't want a concrete state but I never actually finished that refactoring, I'm supportive of this with passing tests

fables-tales (approved):

PR #568, opened 2025-12-10
[prism] Support several one-off node types

reese:
This PR is a bit of a smorgasbord, but all of these node types are fairly isolated in terms of implementation, and given the large number of node types, I figured I'd try and pare down the PRs.

This supports the `alias`, `defined?` `undef`, `BEGIN { }`, `END { }`, `break`, and rational number node types in the Prism implementation and gets 9 more fixtures passing.


fixtures/small/breaks_actual.rb
```diff
@@ -1,2 +1,4 @@
-break no
-break
+loop do
```
reese, line 1: Prism considers `break` outside of a block to be a syntax error, hence this change.

fables-tales: this causes me psychic pain

fables-tales (approved):

froydnj (approved):

PR #574, opened 2025-12-10
[prism] Support inline and multiline versions of `if`/`unless`

reese:
On the tin -- this supports the funky if/unless logic in the Prism implementation. For context, this is complicated because when inline versions of `if` and `unless` get too long, we transform them into the full multiline constructs. (There'll be a similar change later that's similar with `while` and `until`, which have their own idiosyncrasies.)


tests/fixtures_test.rs
```diff
@@ -40,7 +40,6 @@ const DISABLED_PRISM_TESTS: &[&'static str] = &[
     "large_concurrent-ruby_mutex_atomic",
     "large_concurrent-ruby_non_concurrent_map_backend",
     "large_concurrent-ruby_non_concurrent_priority_queue",
-    "large_concurrent-ruby_numeric_cas_wrapper",
```
froydnj, line 43: Whoa!

froydnj (approved):

PR #576, opened 2025-12-10
[prism] Support attribute writes

reese:
This adds support for attribute writes, such as `foo.bar = "baz"`. Note that in these constructs the `method_name` would be `bar=`, so we need to use the `message_loc` to get `bar`. Also somewhat counterintuitively, since it's modeled as a method call, the right hand side is an `ArgumentsNode`, hence the `format_arguments_node` call.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1352,9 +1352,18 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
 
     if skip_receiver || call_node.receiver().is_none() {
         if !is_aref {
+            let method_ident = if call_node.is_attribute_write() {
+                loc_to_string(
+                    call_node
+                        .message_loc()
+                        .expect("Attribute writes must have a message"),
+                )
+            } else {
+                method_name
+            };
             handle_string_at_offset(
                 ps,
-                method_name,
+                method_ident,
                 call_node.message_loc().unwrap().start_offset(),
             );
         }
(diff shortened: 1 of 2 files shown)


froydnj (approved):
No argument from me here, but does this code have to special case `[]=` as well?

reese:
It will yeah, but arefs are going to be a bit more of a pain, so I was going to try and get some quick wins and unblock other things and then tackle them later.

Aref assigns are a lot more complex because we'll need to figure out how to extract which arguments are in the aref and which are the values on the right hand side. For example, if you do `foo.bar[1, 2] = 3`, what you get is a call node with a message name of `[]=` and then an argument list of nodes for 1, 2, and 3, so we'll have to do some weird thing with offsets to figure out which arguments are in the brackets and which ones are the values.

```
              ├── name: :[]=
              ├── message_loc: (1,7)-(1,13) = "[1, 2]"
              ├── opening_loc: (1,7)-(1,8) = "["
              ├── arguments:
              │   @ ArgumentsNode (location: (1,8)-(1,17))
              │   ├── flags: ∅
              │   └── arguments: (length: 3)
              │       ├── @ IntegerNode (location: (1,8)-(1,9))
              │       │   └── flags: decimal
              │       ├── @ IntegerNode (location: (1,11)-(1,12))
              │       │   └── flags: decimal
              │       └── @ IntegerNode (location: (1,16)-(1,17))
              │           └── flags: decimal
              ├── closing_loc: (1,12)-(1,13) = "]"
```

I'm sure that will be a blast, but it's a lot more than I wanted to tackle here.

PR #581, opened 2025-12-11
[chore] Bump `actions/checkout` to latest version

reese:
On the tin -- `actions/checkout@v1/v2` haven't been updated for many years, and while I doubt it has a huge effect on us, it's good to keep up to date with the currently-supported actions.


froydnj (approved):
🧹

PR #582, opened 2025-12-11
[prism] Fix comment rendering after last call arg

reese:
This fixes a small bug in the Prism call comment rendering by making sure that we wind forward all the way to the closing loc of the call, which in most cases is the closing paren. This ensures that we appropriately render all the comments in between the last argument and the closing paren.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1482,7 +1482,7 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
                                     format_block_argument_node(ps, block_argument_node);
                                 }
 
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1481,6 +1481,11 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
                                     }
                                     format_block_argument_node(ps, block_argument_node);
                                 }
+
+                                // Ensure that we render comments between the last argument and closign parens
```
froydnj:
```suggestion
                                // Ensure that we render comments between the last argument and closing parens
```

froydnj (approved):

PR #583, opened 2025-12-11
[prism] Add support for optionally-parenthesized methods

reese:
This re-implements the optional-parentheses logic for Prism. This covers all the places where we we either consider parens optional (like class-level DSLs) or where parens may be meaningful (such as when a name is both a local variable and a method).

This is duplicated pretty much exactly the same as the Ripper version -- we're already re-implementing essentially everything, so I didn't want to try and dry this up by having a method that handles both `CallNode`s and Ripper types, which I think would be pretty crufty.


tests/fixtures_test.rs
```diff
@@ -37,14 +37,10 @@ const DISABLED_PRISM_TESTS: &[&'static str] = &[
...
     "large_concurrent-ruby_ivar",
     "large_concurrent-ruby_java_non_concurrent_priority_queue",
-    "large_concurrent-ruby_mutex_atomic",
     "large_concurrent-ruby_non_concurrent_map_backend",
-    "large_concurrent-ruby_non_concurrent_priority_queue",
     "large_concurrent-ruby_ruby_non_concurrent_priority_queue",
-    "large_concurrent-ruby_synchronized_map_backend",
     "large_concurrent-ruby_truffleruby_map_backend",
     "large_concurrent_ruby_future",
-    "large_dqt",
```
reese, line 47: 🎉

froydnj (approved):

PR #587, opened 2025-12-14
Use `Cow` for some static string values

reese:
Many keywords, operators, delimiters, and so on in `rubyfmt` are static strings, but all throughout the codebase we heap-allocate them with `to_string` a bit willy-nilly, and these are often later cloned, creating more allocations etc. I think we've mostly just used `String` out of inertia, but it's probably unnecessary for most call sites. In reality, a ton of these could eventually should just use `&str`, and `String` should be limited to dynamically-sized use cases like breakables. 

This PR is just the first step in reducing a whole host of these allocations by making the return type of `into_ruby()` be a `Cow<'static, str>` instead of a `String` and converting the easiest ones (keywords for which we already have special tokens, like `def`, `do`, `class`, etc.) to use `Cow::Borrowed`.

In later PRs, I'll convert even more call sites to use `&str` instead of `String`, like delimiters and such which are extremely common in most files. However, even just this PR, by a very rough allocation check, reduces the number of total allocations by something like 3% (for the Ripper implementation on `fixtures/large/concurrent-ruby/ruby_non_concurrent_priority_queue_actual.rb`), implying that we could meaningfully reduce those allocations even more down the line.

### DIFF
--- a/librubyfmt/src/line_tokens.rs
+++ b/librubyfmt/src/line_tokens.rs
@@ -3,6 +3,7 @@ use crate::render_targets::{
     AbstractTokenTarget, BreakableCallChainEntry, BreakableEntry, ConvertType,
 };
 use crate::types::ColNumber;
+use std::borrow::Cow;
 
 pub fn cltats_hard_newline() -> ConcreteLineTokenAndTargets {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Yes!

PR #590, opened 2025-12-14
Don't `Box` closures for parser state methods

reese:
`Box`ing closures has a runtime overhead, since it prevents some compiler optimizations and heap-allocates all of these closures. It's also very verbose and causes the formatted output to be far less intuitive. Now that @froydnj removed the trait implementation that first introduced this in #565, we can :knife: the `Box` usage as well.

This PR is hefty but quite mechanical -- the meat of the change is in `librubyfmt/src/parser_state.rs`, and everything else is just mechanical changes to callsites where I removed `Box::new(` from everything. I'd recommend viewing with whitespace changes hidden.


froydnj (approved):
Yes!

PR #595, opened 2025-12-14
[prism] Handle percent arrays

reese:
This adds support for percent array literals, such as `%w[]` etc. Just one dedicated test for it, but it unblocks a few other test cases that have them in their fixtures.


froydnj (approved):
The test suite could definitely use a little more coverage in this area.

PR #599, opened 2025-12-14
[prism] Keep existing behavior around yield with keyword hash args

reese:
Honestly, I don't remember the context of _why_ we did it this way, but in the Ripper implementation we [made a special-case](https://github.com/fables-tales/rubyfmt/commit/9e5d8cbf43a1b3adefc5c9cba2f8e5da41ca83da) for `yield` calls with arguments that are a single keyword node, e.g. `yield a: b`. I don't offhand remember what this was intended to prevent, but it was clearly intentional we put it there and I don't want to get Chesterton's fenced on some obscure yield behavior, so this adds that check in the Prism implementation too.


froydnj (approved):
Consistency ✅

PR #605, opened 2025-12-15
[prism] Fix block spacing with lonely operators in call chains

reese:
This gets `test_prism_small_method_chains` passing by fixing several small issue. The main one was a small issue with lonely operators after blocks in call chains. We have special handling for the dot operators in calls, but for whatever reason the machinery was only checking for `Dot` and `DirectPart { "&." }` instead of the `LonelyOperator` token (which is equivalent to the latter). This adds the appropriate check and also makes a small adjustment to ensure we render comments correctly, as well as making sure that call chains that start with parentheses node multiline correctly.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1707,6 +1707,7 @@ fn format_call_chain(ps: &mut ParserState, call_node: ruby_prism::CallNode<'_>) 
                     }
                 });
                 ps.end_indent_for_call_chain();
+                ps.shift_comments();
             },
         );
     });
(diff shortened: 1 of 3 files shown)


froydnj (approved):

librubyfmt/src/format_prism.rs
```diff
@@ -1727,6 +1728,7 @@ fn call_chain_elements_are_user_multilined(
             prism::Node::CallNode { .. }
                 | prism::Node::ConstantReadNode { .. }
                 | prism::Node::ConstantPathNode { .. }
+                | prism::Node::ParenthesesNode { .. }
```
reese, line 1731: I'm not sure if this is accidental or intended in the Ripper implementation, but we check for leading literals (like arrays etc.) by [checking for specific Expression types](https://github.com/fables-tales/rubyfmt/blob/b2b28df470fd918b0ac667829d301a7d1032151a/librubyfmt/src/render_targets.rs#L384-L386), but in the Ripper call chain node types, [`Paren` is separate from `Expression`](https://github.com/fables-tales/rubyfmt/blob/b2b28df470fd918b0ac667829d301a7d1032151a/librubyfmt/src/ripper_tree_types.rs#L1719-L1720), so it is not considered a "leading expression".

PR #607, opened 2025-12-15
[prism] Properly enforce hash rocket usage rules

reese:
The intended outcome is that if all keys are symbols, we use symbols, but if even one uses a hash rocket, we render everything as hash rockets. The tricky bit here is that symbols often include their trailing colon (and if we're converting to a hash rocket, need to be given a leading colon), so this PR adds all the logic to trim or add colons where necessary when converting between key types.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1848,7 +1848,7 @@ fn format_assoc_node(ps: &mut ParserState, assoc_node: prism::AssocNode) {
             ps.emit_ident(":".to_string());
         }
 
-        format_node(ps, dbg!(assoc_node.key()));
+        format_node(ps, assoc_node.key());
         if as_symbol {
             ps.emit_ident(":".to_string());
         } else {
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2499,15 +2511,29 @@ fn format_hash_node(ps: &mut ParserState, hash_node: prism::HashNode) {
...
-            ps.breakable_of(BreakableDelims::for_hash(), |ps| {
-                ps.emit_soft_indent();
-                format_list_like_thing(
-                    ps,
-                    hash_node.elements(),
-                    hash_node.closing_loc().end_offset(),
-                    false,
-                );
-                ps.wind_dumping_comments_until_offset(hash_node.closing_loc().end_offset());
+            let all_symbol_keys = dbg!(&hash_node)
```
froydnj: Can we remove the debugging code before merge?

froydnj (approved):

PR #611, opened 2025-12-15
Convert `ConcreteLineToken::Op` to use `Cow`

reese:
On the tin. Some op tokens are copied from their respective nodes, so I opted to have this take a `Cow` itself instead of trying to wrangle lifetimes for the time being, allowing most nodes to pass a `&str` and the few operator writes to continue using `String`.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use ruby_prism as prism;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2971,7 +2969,9 @@ fn format_constant_path_write_node(
     constant_path_write_node: prism::ConstantPathWriteNode,
 ) {
     format_constant_path_node(ps, constant_path_write_node.target());
-    ps.emit_op(" = ".to_string());
+    ps.emit_space();
+    ps.emit_op(Cow::Borrowed("="));
+    ps.emit_space();
```
froydnj, line 2974: Why the changes here and elsewhere to emit spaces around the operator along with this change?  Are we just bringing that all by for convenience, or are there formatting reasons why we want this (more correct line-breaking?)?

froydnj (approved):

librubyfmt/src/format_prism.rs
reese: Just for consistency -- in some places we do like " op " and others we explicitly emit spaces. I don't think we ever do line breaks off of spaces, but if they ever did become relevant, I'd rather have all the same kind than a mixture of strings and tokens.

PR #613, opened 2025-12-15
[ripper] Correctly break breakables with multiline double-quote strings

reese:
While looking at Prism failures, I noticed that this Ripper output actually seemed wrong. Generally, the rule for breakables is that if the contents of the breakable are on multiple lines, then the breakable is also multiline. The Prism implementation did this correctly for strings, but the Ripper implementation didn't.

For double-quoted strings with newline characters, we adjust the `current_line_number` in the parser state, but incorrectly failed to update that for the breakable stack. In most cases, callers that do this will use heredocs (and the heredoc machinery correctly handles this case), but on the rare usage of a multiline double-quoted string, it didn't.

In short:

```ruby
# before
method("string on
multiple lines")

# after
method(
  "string on
multiple lines"
)
```

### DIFF
--- a/fixtures/small/gemfile_expected.rb
+++ b/fixtures/small/gemfile_expected.rb
@@ -17,8 +17,10 @@ scope(
   :with_ticket_types,
   -> {
     group("shops.id")
-      .joins("LEFT JOIN foos ON foos.bar_id
-= bars.id")
+      .joins(
+        "LEFT JOIN foos ON foos.bar_id
+= bars.id"
+      )
       .select("foos.*, COUNT(bars.id) count")
   }
 )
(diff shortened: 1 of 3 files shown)


froydnj (approved):
I am unexcited about this causing formatting changes on Stripe's codebase, but it does seem like the right thing to do.

reese:
I'd bet that the number of changes in the codebase should be ~minimal. My memory from rolling rubyfmt out in the Stripe codebase was that the number of non-heredoc multiline strings was pretty small. (They also formatted in less-than-pleasant ways generally and I think were usually accidents, so I think that discouraged them from proliferating too much.)

froydnj:
> 'd bet that the number of changes in the codebase should be ~minimal.

FWIW, we rolled this out and it wound up changing small dozens of files.  So yeah, not that bad!

PR #625, opened 2025-12-17
[prism] Correctly handle spacing between `end` and `def` tokens

reese:
Normally, we add an extra newline after `end` nodes, but we have special handling for "method annotations" (mostly for `sig` and friends) where we detect `end`s followed by a `def` and don't insert this newline. This was previously busted because we were skipping emitting `AfterCallChain` tokens for single-element block calls (that is, calls with blocks not part of a call chain), which this handling depends on.

While investigating this, I also fixed a long-standing bug! It's nothing user-facing, but we had a set of tests that perennially failed in `debug` due to some assertions failing. The assertions failed because when we insert newlines in the render queue, we were failing to update the `index_of_last_hard_newline` attribute, which then triggered some, er, [_descriptive_](https://github.com/fables-tales/rubyfmt/blob/d033a7b24fe413490ee213a167b2a9559f2b8a06/librubyfmt/src/intermediary.rs#L272) debug assertions. This is now fixed!

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1058,7 +1058,7 @@ fn format_module_node(ps: &mut ParserState, module_node: prism::ModuleNode) {
 }
 
 fn format_def_node(ps: &mut ParserState, def_node: prism::DefNode) {
-    ps.emit_keyword("def");
+    ps.emit_def_keyword();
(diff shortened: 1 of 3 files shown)


froydnj (approved):
"descriptive" 😆 

🎉

PR #626, opened 2025-12-18
[prism] Properly handle single ParenNode arguments

reese:
At some point, we decided that args that are just a single expression wrapped in parens should e unwrapped so that calls like `a (1)` become `a(1)`. I'll say for the record I think this is a bit of a weird precedent IMO and I don't remember why this was decided on, but for the time being I'll keep this the same as in Ripper and implement it in Prism.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1448,7 +1448,7 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
         let has_only_empty_paren_arg = !is_aref
             && !is_aref_write
             && call_node.arguments().is_some_and(|args| {
-                args.arguments().iter().count() == 1
+                args.arguments().len() == 1
                     && is_empty_parentheses_node(&args.arguments().iter().next().unwrap())
             });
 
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1443,7 +1443,18 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
                 start_loc_for_call_node_in_chain(&call_node),
             );
         }
-        if let Some(arguments) = call_node.arguments() {
+
+        // Calls with only empty parens (`foo ()`) are treated as calls without args
+        let has_only_empty_paren_arg = !is_aref
+            && !is_aref_write
+            && call_node.arguments().is_some_and(|args| {
+                args.arguments().iter().count() == 1
```
froydnj: Maybe land this PR after the `ruby-prism` 1.7 bump so this can be `len()`?

librubyfmt/src/format_prism.rs
```diff
@@ -1508,10 +1519,35 @@ fn format_call_node(ps: &mut ParserState, call_node: prism::CallNode, skip_recei
                 } else {
                     BreakableDelims::for_kw()
                 };
+
+                // Check if we can unwrap a single parenthesized argument when we're adding
+                // method call parens. This handles cases like `a (1)` -> `a(1)` where the
+                // parens around `1` were just for argument grouping, not expression grouping.
+                let maybe_unwrapped_single_arg = if should_use_parens
+                    && arguments.arguments().iter().count() == 1
```
froydnj: Here too.

froydnj (approved):

PR #627, opened 2025-12-18
[prism] Bump ruby-prism to 1.7.0

reese:
On the tin -- the main thing I want in this version is @froydnj's change to add `.len()` to `NodeList` and `ConstantList` 🎉 

So this PR just bumps the version and updates a bunch of `.iter().count()` to `.len()`.


froydnj (approved):
Literally pushed this branch like 30 seconds before this PR arrived in my inbox: https://github.com/fables-tales/rubyfmt/compare/froydnj-update-ruby-prism?expand=1

Still not fast enough to beat @reese !

PR #628, opened 2025-12-18
[prism] Remove node_list_is_empty :broom:

reese:
In the new prism version, `is_empty` is available to node and constant lists, so we don't need this helper anymore.


froydnj (approved):
🎉

PR #633, opened 2025-12-19
[prism] Fix indentation inside arefs at the start of call chains

reese:
In Ripper, arefs are represented as a standalone expression type, so in call chains where the first node was an `Aref` node, we would format the whole thing first before emitting a `BeginCallChainIndent` token. In Prism, arefs are separate calls, so the first element in `Array[1, 2]` would just be the constant node for `Array`.

This PR adds some handling for aref calls at the top of chains to explicitly render all leading arefs before starting the call chain indentation.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1815,6 +1815,23 @@ fn format_call_chain(ps: &mut ParserState, call_node: ruby_prism::CallNode<'_>) 
                 // must be additional calls -- you cannot insert literals into call chains
                 let first_expression = call_chain_elements.remove(0);
                 format_node(ps, first_expression);
+
+                // Arefs like `Array[1, 2]` are represented as a call chain where
+                // the receiver is the constant name `Array` and the aref is a separate call node.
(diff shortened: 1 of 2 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -1815,6 +1815,23 @@ fn format_call_chain(ps: &mut ParserState, call_node: ruby_prism::CallNode<'_>)
...
+                // the receiver is the constant name `Array` and the aref is a separate call node.
+                // However, we don't want to start the call chain indent until after the aref, since the
+                // aref is "part of" the first expression.
+                // We loop here to handle chained arefs like `matrix[0][1].foo`.
+                while let Some(next) = call_chain_elements.first() {
+                    if let Some(call_node) = next.as_call_node()
+                        && call_node.call_operator_loc().is_none()
+                    {
+                        let aref = call_chain_elements.remove(0);
+                        format_call_node(ps, aref.as_call_node().unwrap(), true);
```
froydnj: We can just use `call_node` here and avoid recasting, couldn't we?

froydnj (approved):

PR #634, opened 2025-12-19
[prism] Fix rendering nested heredocs

reese:
On the tin -- the code that handles rendering heredocs in interpolations was already run for `InterpolatedStringNode` but in our test also needs to run for `EmbeddedStatementsNode`s, since in our cursed fixture we have a heredoc with an interpolation (the former) but the interpolation itself also has a heredoc in it (the latter).


froydnj (approved):
Getting down to the heredoc corner cases!

PR #635, opened 2025-12-19
[prism] Handle %Q string nodes

reese:
On the tin -- this updates the string escaping bits to transform `%Q` into double-quoted strings.

### DIFF
--- a/librubyfmt/src/string_escape.rs
+++ b/librubyfmt/src/string_escape.rs
@@ -17,7 +17,7 @@ pub fn single_to_double_quoted(content: String, start_delim: &str, end_delim: &s
         };
 
         let regexp = Regex::new(&format!(
-            r#"(?<!\\)(\\\\)*(\"|\\\\{}|\\\\{})"#,
+            r#"(?<!\\)(\\\\)*(\"|\\{}|\\{})"#,
             fancy_regex::escape(start_delim),
             fancy_regex::escape(end_delim)
         ))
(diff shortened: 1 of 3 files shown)


librubyfmt/src/string_escape.rs
```diff
@@ -17,7 +17,7 @@ pub fn single_to_double_quoted(content: String, start_delim: &str, end_delim: &s
         };
 
         let regexp = Regex::new(&format!(
-            r#"(?<!\\)(\\\\)*(\"|\\\\{}|\\\\{})"#,
+            r#"(?<!\\)(\\\\)*(\"|\\{}|\\{})"#,
```
reese, line 20: Not sure why I wrote it with these extra escapes originally, the updated version is the same as [what we used in Ripper](https://github.com/fables-tales/rubyfmt/blob/56cec999d663d1adc665dbec2ace2c8f152dae8b/librubyfmt/rubyfmt_lib.rb#L567-L568).

froydnj (approved):
Man, I feel like the string escape bits could be improved for efficiency at some point.

reese:
> Man, I feel like the string escape bits could be improved for efficiency at some point.

100%. One thing to consider too is that a lot of the reason it's overly complex (and that I'll probably have to shove even more random escaping stuff in there in another PR soon) is that we have to reimplement so much random behavior from `String#inspect` to match Ripper. At some point, we should probably consider reconsidering how much of that we actually want vs how much we can just let Prism escape for us.

PR #638, opened 2025-12-19
[prism] Fix comment rendering for multiline brace blocks

reese:
Brace blocks are weird for a number of reasons, not the least of which is that they have a lot of strange newline usages to be able to render expressions either on a single line `.map { foo }` or on multiple lines like a do/end block. To properly render comments in the multiline form, we generally need to use soft newlines, which are special cased inside the comment rendering machinery to insert at the right index. Previously I had them using hard newlines (automatically inserted in `format_node`, but this instead updates it to manually insert the appropriate soft newlines/indents so that comments render in the right spots.

### DIFF
--- a/tests/fixtures_test.rs
+++ b/tests/fixtures_test.rs
@@ -18,10 +18,7 @@ const DISABLED_RIPPER_TESTS: &[&'static str] = &[
 ];
 
 // These tests will have their prism variants automatically ignored
-const DISABLED_PRISM_TESTS: &[&'static str] = &[
-    "small_brace_blocks_with_no_args",
-    "small_dyna_symbol_with_escapes",
-];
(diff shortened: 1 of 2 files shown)


tests/fixtures_test.rs
```diff
@@ -18,10 +18,7 @@ const DISABLED_RIPPER_TESTS: &[&'static str] = &[
 ];
 
 // These tests will have their prism variants automatically ignored
-const DISABLED_PRISM_TESTS: &[&'static str] = &[
-    "small_brace_blocks_with_no_args",
-    "small_dyna_symbol_with_escapes",
-];
```
froydnj, line 24: Was super excited to see this until I noticed it just got reformatted 😂

froydnj (approved):

tests/fixtures_test.rs
reese: The last one will be coming soon!

PR #640, opened 2025-12-19
[prism] Manually handle heredoc inner string formatting to preserve escapes

reese:
On the bright side, this fixes the last existing fixture test for Prism <> Ripper compatibility! :tada:

On the down side, it's a giant PITA. In short, it reworks heredoc escaping where we essentially do all of the escaping ourselves by hand instead of offloading it to Ripper, which allows to maintain some existing features around escaping.

---

Full details:

### Escape Sequences

In heredocs, there are cases where users may manually insert escape sequences like `\n`, but Ripper sees that and then returns it in its `unescaped` string value as an actual newline. That is, it makes not distinction between `\\n` and a newline character in the source, since at runtime they're the same. In our case, we'd like to keep that distinction.

In order to maintain those two differences, we cannot use the `unescaped` string and need to use the `content_loc` and then clean it ourselves. The main reason this is a PITA is that `unescaped` did a lot of the heavy lifting for us, like trimming leading whitespace for squiggly (`<<~`) heredocs. There's a lot of nuance in that implementation w.r.t. how it interacts with interpolated nodes etc., so this adds a fair amount of complexity. (This is essentially what I was alluding to with [this comment](https://github.com/fables-tales/rubyfmt/pull/635#issuecomment-3675310539) on an earlier issue -- it may be worth reconsidering if this is a valid tradeoff.)

### Converting Unicode characters

This fixture also has a bunch of Japanese kanji in it, which in Ripper is read through `String#inspect` and converted into unicode escape characters. Personally I (and other users -- see #454) think this isn't great, and I think it's worthwhile to keep them as the original characters whenever possible. (Note that in the Stripe codebase, all strings would have already been converted to unicode escapes and would stay that way if reformatted with Prism, so this shouldn't cause a reformat. But it may make some teams happy if you went back and reverted them to the old way!)

To that end, I opted to leave the Prism implementation such that it leaves Kanji as-is. For the tests, I updated the test harness to have Prism-only and Ripper-only tests and split this one test into two explicit ones.


librubyfmt/src/format_prism.rs
```diff
@@ -813,46 +813,122 @@ fn format_heredoc(ps: &mut ParserState, heredoc: HeredocNodeType, heredoc_symbol
...
-            _ => false,
-        };
+    let should_render = match peekable.peek() {
+        Some((_, prism::Node::StringNode { .. })) => loc_to_string(
+            peekable
+                .peek()
+                .unwrap()
+                .1
+                .as_string_node()
+                .unwrap()
```
froydnj: idk if this would be better written as something like `peek().map(|(_, node)| node.as_string_node().map(|sn| loc_to_str(sn.content_loc()).starts_with('\n'))).unwrap_or(false)`

librubyfmt/src/format_prism.rs
```diff
@@ -813,46 +813,122 @@ fn format_heredoc(ps: &mut ParserState, heredoc: HeredocNodeType, heredoc_symbol
...
+    // For squiggly heredocs, calculate the common indentation to strip.
+    // We only look at lines that start at the beginning of a StringNode part
+    // (not continuations after an interpolation on the same line).
+    let common_indent = if heredoc_kind.is_squiggly() {
+        parts
+            .iter()
+            .enumerate()
+            .filter_map(|(i, part)| {
+                if let prism::Node::StringNode { .. } = part {
+                    let content = loc_to_str(part.as_string_node().unwrap().content_loc());
```
froydnj: Minor preference for:

```rust
if let Some(node) = part.as_string_node() {
  let content = loc_to_str(node.content_loc());
  ...
}
```

librubyfmt/src/format_prism.rs
```diff
@@ -813,46 +813,122 @@ fn format_heredoc(ps: &mut ParserState, heredoc: HeredocNodeType, heredoc_symbol
...
+                    // (i.e., if the previous part ended with a newline, or this is the first part)
+                    let prev_ends_with_newline = if i == 0 {
+                        true
+                    } else {
+                        match &parts[i - 1] {
+                            prism::Node::StringNode { .. } => {
+                                loc_to_str(parts[i - 1].as_string_node().unwrap().content_loc())
+                                    .ends_with('\n')
+                            }
+                            _ => false, // After interpolation, might not be at line start
```
froydnj: Same deal here with something like `parts[i - 1].as_string_node().map(|node| loc_to_str(node.content_loc()).ends_with('\n')).unwrap_or(false)`

froydnj (approved):
w00!  Very excited to see how this does on Stripe's codebase once it's merged.

reese:
>  Very excited to see how this does on Stripe's codebase once it's merged.

Yes definitely know what you find! There's still some node types that are `todo!()` that we didn't have fixtures for, so I'm still going through those and implementing them, but hopefully this should get you through a good chunk of the codebase.

PR #652, opened 2025-12-19
[prism] Only treat arefs as calls/assigns when there's no call operator

### LINKED ISSUE #651: [prism] safe navigation on `[]` formats to invalid Ruby
cc @reese 

Testcase:

```ruby
x&.[](y)
```

`rubyfmt --prism` produces:

```ruby
x&.[y]
```

which is not valid syntax.

reese:
Closes #651 

We were checking for arefs by looking for the `[]`/`[]=` message locs, but we also need to look for the operator so we can differentiate between `x[y]` and `x.[](y)`


froydnj (approved):
That was a lot nicer than what I was going to try 😅

PR #660, opened 2025-12-20
[prism] Remove unnecessary node conversions

reese:
This is very much a micro-optimization, but there's several formatting functions that either are implicit nodes and do nothing or that are simple keywords (e.g. `nil`) that don't actually need access to the cast node value. For these node types, we should just not pass the node at all to avoid the (very tiny) overhead of casting them.


froydnj (approved):
This kind of PR is my kind of PR.

froydnj:
My suspicion is that this function might be better written as an if-let chain, so as to avoid all the conversions + unwraps.

I'm not totally sure it would be better, because I _think_ the compiler might be able to optimize the conversions and panicking code away as things stand right now, but I'd want to look at the assembly before and after to be sure .

PR #671, opened 2025-12-20
[prism] Ensure we wind to the closing loc for all call nodes

### LINKED ISSUE #666: [prism] invalid ruby when formatting longish call chains
Input, unchanged by ripper:

```ruby
def validate(params:)
  # run check
  set = [params.percent_off, params.amount_off, params.script]
    .compact
    .filter(
      &:present?
    )
    .count
end
```

`rubyfmt --prism` turns that into:

```ruby
def validate(params:)
  # run check
  set = [params.percent_off, params.amount_off, params.script]
    .compact
    .filter(
      &:present?
    )

    .count
end
```

which the interpreter complains about.
cc @reese

reese:
Closes #666 ([obligatory Iron Maiden](https://youtu.be/WxnN05vOuSM?t=110))

Call chains have some finicky line winding properties since we mostly handle the calls manually instead of passing it through the `format_node` machinery. We should always be winding to the end of each call node to make sure we wind to the closing paren, so this adds that to `format_call_node` to catch any places we forget to wind forward during the chain.

### DIFF
--- a/fixtures/small/method_chains_actual.rb
+++ b/fixtures/small/method_chains_actual.rb
@@ -54,6 +54,12 @@ foo.items.map { p(_1) }.last
 
 hashes.sort_by { |hsh| hsh[:start_time] }.reverse
 
+hash.sort_by(
+    &:max
+  )
+  .keys
+
+
 params(
   route: String,
   config: T.nilable(Some::Really::Long::Type::Name),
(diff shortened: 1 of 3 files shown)


froydnj (approved):
Commenting winding is a bit weird, so I am going to trust that this is correct.

PR #679, opened 2025-12-21
Never strip parens from method call arguments

### LINKED ISSUE #673: [prism] parenthesized expressions are syntactically meaningful in some cases
Ruby:

```ruby
results, errors = Helpers.generate(
  (spec_names.uniq unless spec_names.include?("reject"))
)
```

`rubyfmt --prism` generates:

```
results, errors = Helpers.generate(
  spec_names.uniq unless spec_names.include?("reject")
)
```

which is invalid syntax:

```
-:2: syntax error, unexpected `unless' modifier, expecting ')'
  spec_names.uniq unless spec_names.include?("reject")
-:3: syntax error, unexpected ')'
-: compile error (SyntaxError)
```

So...maybe we do need to parenthesize expressions if they appeared with parens in the source?

cc @reese

reese:
Closes #673

I was doing some spelunking to determine why we remove parens from nodes like `a (1)` in the Ripper implementation. I assumed that this had been an intentional choice and that we'd have some dedicated handling for it in the rendering code, but we don't! The only place that actually strips it is one place that converts from an `AABParen::Paren` to an `Expression` by returning the paren's inner expression value, and given that this was in the comparatively early days of `rubyfmt` and not applied uniformly (e.g. `a((1))` won't strip parens) and isn't safe in every context (e.g. `a (foo unless bar)` becomes `a(foo unless bar)`, which is invalid syntax), I'm led to believe this was just an oversight and not an explicit design choice.

To that end, this PR does what I think is the more correct behavior, which is that it never modifies parens for arguments in any case. This is (1) true to what the user's AST was previously and (2) means we never have to be concerned about when it is or isn't safe to strip them. This applies to both the Ripper and Prism implementations.

(For the Stripe folks, note that this shouldn't cause any changes to codebases that have already been formatted AFAICT, since any places where it's safe to strip parens would already have had them formatted away, so this should be a no-op.)

### DIFF
--- a/fixtures/small/paren_expr_calls_actual.rb
+++ b/fixtures/small/paren_expr_calls_actual.rb
@@ -14,3 +14,5 @@ other_cool_method (a + b).round(4)
   )
   &.rules)
   .freeze
+
+beekeep (the_bees unless not_the_bees!)
--- a/fixtures/small/paren_expr_calls_expected.rb
+++ b/fixtures/small/paren_expr_calls_expected.rb
@@ -1,4 +1,4 @@
-a(1)
+a((1))
 other_cool_method((a + b).round(4))
 
 # rubocop:disable PrisonGuard/PrivateModule
@@ -16,3 +16,5 @@ other_cool_method((a + b).round(4))
   )
   &.rules)
   .freeze
+
+beekeep((the_bees unless not_the_bees!))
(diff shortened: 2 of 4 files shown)


fixtures/small/paren_expr_calls_actual.rb
```diff
@@ -14,3 +14,5 @@
   )
   &.rules)
   .freeze
+
+beekeep (the_bees unless not_the_bees!)
```
froydnj, line 18: That the whitespace is significant here is just 🤯 .

froydnj (approved):

Issue #417, opened 2023-02-22
rubyfmt ignores `embdoc` comments (`=begin` and `=end` comments)


reese:
* Rubyfmt git sha: 7f56694aa5c683b333bfe30e545eac02de6009ae 

## Input file

```ruby
=begin
  Bees, bees, not the bees!
=end
```

## Rubyfmt's output

```

```

## This breaks because

These comments should stick around! I don't think many people use them, but we should probably support them anyways. If anyone is interested in doing this, there are some docs [here](https://github.com/kddnewton/ripper-docs/blob/main/events.md#embdoc) on how this is handled in Ripper.

mcmire:
+1 on including support for these style of comments. I often use these as a quick way of commenting out large blocks of code.

reese:
`v0.12.0` has this working now! [Example](https://rubyfmt.run/#PWJlZ2luCiAgQmVlcywgYmVlcywgbm90IHRoZSBiZWVzIQo9ZW5kCgojIEFuZCByZWd1bGFyIGNvbW1lbnRzIHRvbyEK).

PR #683, opened 2025-12-21
[prism] Support embdoc (`=begin`/`=end`) comments

reese:
Related to #417 

The Ripper implementation just ignores embdocs, but Prism reports them the same as other comments, so it's not a huge lift to support them -- it's mostly a matter of updating the comment machinery to not assume that comments are always single-line and might be multiple lines long.

### DIFF
--- a/fixtures/small/prism/embdoc_actual.rb
+++ b/fixtures/small/prism/embdoc_actual.rb
@@ -1,0 +1,43 @@
+def haiku
+  # Violets -
+    how_precious
+    on_a_mountain_path
+
+=begin
+    Early autumn -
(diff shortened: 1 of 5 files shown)


fixtures/small/prism/embdoc_actual.rb
```diff
@@ -0,0 +1,43 @@
...
+def
+=begin
+=end
+a; end
+
+:\
+=begin
+not the symbol name
+=end
+symbol_name
```
reese, line 43: Fun Ruby syntax discovery of the day

fables-tales (approved):

PR #686, opened 2025-12-21
[prism] Implement remaining pattern matching nodes

reese:
On the tin -- this means we'll have implemented every node that Prism emits, so no more `todo!()`!

I'm open to bikeshedding on the specifics, but in general I modeled these nodes after existing nodes elsewhere, so array patterns are formatted like arrays, hash patterns like hashes, etc. For multiline arrays/hashes, they emit one indentation level up, which is the same as what we do for conditionals. Everything else I think more or less straightforward. There's quite a lot of fixtures, so that's probably a great place to start.


froydnj (approved):
These all seem reasonable to me!

Issue #459, opened 2024-01-17
not(var.nil?) gets wrongfully converted to not var.nil? which can lead to invalid ruby code


onvbqalc:
* Ruby version: `3.1.2`
* Rubyfmt git sha: `e00d2ab89fd4b0b85a7897fac393c1ad987136de`
```
==> rubyfmt: stable 0.10.0 (bottled), HEAD
Ruby autoformatter
https://github.com/fables-tales/rubyfmt
/opt/homebrew/Cellar/rubyfmt/0.10.0 (8 files, 15.4MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-01-07 at 04:41:02
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/r/rubyfmt.rb
```

## Input file

```ruby
class MyClass
  def my_method
    a.nil? || not(b.nil?) || c.nil?
  end
end
```

## Rubyfmt's output

```
# First format - no output but brackets () are removed
# leaving conditional as: a.nil? || not b.nil? || c.nil?
#
# Second format:
Rubyfmt detected a syntax error in the ruby code being executed
```

## This breaks because
Code should have been left as-is on the first format - instead, it got converted to invalid ruby code:
```
a.nil? || not b.nil? || c.nil?
```
(is not valid ruby)

reese:
`v0.12.0` (the latest release) has the fix for this. [Example](https://rubyfmt.run/#Y2xhc3MgTXlDbGFzcwogIGRlZiBteV9tZXRob2QKICAgIGEubmlsPyB8fCBub3QoYi5uaWw/KSB8fCBjLm5pbD8KICBlbmQKZW5k).

Thanks for the report!

PR #687, opened 2025-12-22
[prism] Support `not` with parens

reese:
Part of #459 

`not` sometimes has semantically-meaningful parentheses (such as in `not(foo) || bar`), so this PR just leaves them as-is if they originally had parens. In other similar cases we might do things like use `ps.current_formatting_context_requires_parens()`, but I think `not` has additional cases that that helper doesn't cover (like in the case of `not(foo).class` or some other strange usage), so I opted to have it just always use parens if it had them before.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1944,13 +1944,25 @@ fn format_call_node(
 }
 
 fn format_unary_operator(ps: &mut ParserState, call_node: prism::CallNode, method_name: String) {
+    // We need to preserve parens for `not`, they can be semantically meaningful
+    let is_not_with_parens = method_name == "!"
+        && call_node
+            .message_loc()
+            .map(|loc| loc_to_string(loc) == "not")
+            .unwrap_or(false)
+        && call_node.opening_loc().is_some();
+
     let operator_symbol = match method_name.as_str() {
         "!" => {
             // `not` and `!` both have a `name` of `!` but different messages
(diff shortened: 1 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2947,9 +2964,18 @@ fn split_node_into_call_chains(node: prism::Node) -> Vec<Vec<prism::Node>> {
...
     while let Some(receiver) = maybe_receiver {
-        maybe_receiver = receiver
-            .as_call_node()
-            .and_then(|call_node| call_node.receiver());
+        maybe_receiver = receiver.as_call_node().and_then(|call_node| {
+            // Don't traverse into unary operators, they should not be treated as part of a call chain.
+            let method_name = const_to_string(call_node.name());
+            let is_unary_operator = call_node.arguments().is_none()
+                && call_node.call_operator_loc().is_none()
+                && matches!(method_name.as_str(), "-@" | "+@" | "!" | "~");
```
froydnj, line 2972: Do we have tests for this?  I think this is correct behavior, but it would be nice if it was covered somewhere.

froydnj (approved):

librubyfmt/src/format_prism.rs
reese: This is what `not(false).class` and `foo.not(bar)` in the fixtures I added are intended to cover. (The former is the main bit where this is relevant -- without this change, you get `false!.class`.)

PR #688, opened 2025-12-22
Avoid cloning `FormattingContext` on every breakable

reese:
We clone the `Vec<FormattingContext>` on every single breakable, but we never modify it and just do a pretty simple check with it later -- all we need to know is whether there's a specific context type in that vec, which we can precompute at the breakable's initialization and avoid a clone.


froydnj (approved):
Love it.

PR #690, opened 2025-12-22
Do more efficient multiline tracking for breakables

reese:
Currently, breakables maintain a `HashSet` of every line they visit. The use case for this is that breakables want to know if they've visited multiple lines, so they do `hash_set.len() > 1`. For this use case, a full hash set is overkill -- we have to heap allocate the set, hash all the entries, etc. when we can do something much simpler: store the first line and then check if we ever visit a different one.

### DIFF
--- a/librubyfmt/src/render_targets.rs
+++ b/librubyfmt/src/render_targets.rs
@@ -3,7 +3,6 @@ use crate::line_tokens::{AbstractLineToken, ConcreteLineToken, ConcreteLineToken
 use crate::parser_state::FormattingContext;
 use crate::ripper_tree_types::CallChainElement;
 use crate::types::LineNumber;
-use std::collections::HashSet;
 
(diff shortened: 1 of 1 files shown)


librubyfmt/src/render_targets.rs
```diff
@@ -218,7 +217,6 @@ pub enum MultilineHandling {
 #[derive(Debug, Clone)]
 pub struct BreakableCallChainEntry {
     tokens: Vec<AbstractLineToken>,
-    line_numbers: HashSet<LineNumber>,
```
reese, line 221: Call chains base multilining on the line numbers of the nodes, not the ones reported during rendering, so these were written but never read 🔪

froydnj (approved):
Nice!

Just wanted to say -- in testing, it is very noticeable how much faster prism formatting is vs. ripper, and I'm very excited to see these sorts of fundamental algorithmic improvements go in and add to how much faster prism formatting is.  Thank you!

PR #691, opened 2025-12-22
Short-circuit before rendering line length checks for breakables

reese:
`single_line_string_length` for both types of breakables is a very expensive call. I'd like to optimize it further, but at the moment essentially clones every token and renders it into a single-line string to see if that string is over 120 characters. This also the very first thing we do on every single render of any breakable, even though there are other conditions we could check to completely obviate it.

This PR delays the `single_line_string_length` until after we do a few other much cheaper checks. I think the main place this will help is when code has already been formatted, since breakables will already be on multiple lines in the source and thus `breakable.is_multiline()` will probably be a simple boolean check and skip the entire preliminary rendering.

### DIFF
--- a/librubyfmt/src/render_queue_writer.rs
+++ b/librubyfmt/src/render_queue_writer.rs
@@ -224,14 +224,15 @@ impl RenderQueueWriter {
     }
 
     fn format_breakable_entry(accum: &mut Intermediary, be: BreakableEntry) {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
I think the relevant function could also be improved by not accumulating into a single string, but summing the intermediary string lengths.

PR #711, opened 2025-12-23
[prism] Wrap calls with blocks in `BreakableCallChainEntry` for correct line length handling

### LINKED ISSUE #701: [prism] different breaking with block calls
I don't understand why this happens, exactly, but...input, unchanged by ripper:

```ruby
module ODM
  class Model
    def self.load_all_by_into_hash(keys, extra = {}, options = {})
      hash = {}
      Util::Callsite.capture_query_callsite(collection: self.collection_name, allow_disabling: true) do
        for_each_by_key(keys, extra, options) do |obj|
          hash[obj.key] = obj
        end
      end

      hash
    end
  end
end
```

`rubyfmt --prism` wants to break the call like so:

```ruby
module ODM
  class Model
    def self.load_all_by_into_hash(keys, extra = {}, options = {})
      hash = {}
      Util::Callsite
        .capture_query_callsite(collection: self.collection_name, allow_disabling: true) do
          for_each_by_key(keys, extra, options) do |obj|
            hash[obj.key] = obj
          end
        end

      hash
    end
  end
end
```

which causes a lot of changes on an already-formatted codebase.  Over-eager breaking can be also seen in calls like `model_class.props.each do |key, rules|`, where the calls each get formatted onto their own line, despite not being anywhere near the line length limit.  But the weird thing is that these don't seem to happen consistently.

cc @reese

reese:
Closes #701 

When doing line-length calculations for block calls, we have some machinery that tries to ignore the contents of the block when calculating the length, essentially popping tokens until it gets back to the block params. This is so that multiline blocks at the end of a chain (like `Klass.do_some_stuff_in_a_block do |params| #...`) are only considered insofar as their line length affects the line of the call chain itself, meaning they won't for example break the whole chain if just the block contents go over the line length.

To do this, [this bit](https://github.com/fables-tales/rubyfmt/blob/933408494df678f2388e98262ad18f741bef822b/librubyfmt/src/render_targets.rs#L269-L303) handles this token-popping, and it assumes that blocks are always wrapped in a breakable entry. For brace blocks, this worked since the brace block is _itself_ a breakable, but in Ripper the entire block call would also have been wrapped in a `BreakableCallChainEntry` because that's [how `method_add_arg` blocks were handled](https://github.com/fables-tales/rubyfmt/blob/933408494df678f2388e98262ad18f741bef822b/librubyfmt/src/format.rs#L2600).

The prism implementation didn't do this before, so when calculating the line length of a block call with nested blocks inside of it, it was only unwinding until it saw block params, which would be inside the inner block and not the outer one. This caused some strange multilining that could seem ~random, since it only affected certain arrangements of block calls.

### DIFF
--- a/fixtures/small/nested_block_breakable_actual.rb
+++ b/fixtures/small/nested_block_breakable_actual.rb
@@ -1,0 +1,4 @@
+Util::Callsite.capture_query_callsite(collection: self.collection_name, allow_disabling: true) do
+  for_each_this do |obj|
+  end
+end
--- a/fixtures/small/nested_block_breakable_expected.rb
+++ b/fixtures/small/nested_block_breakable_expected.rb
@@ -1,0 +1,4 @@
+Util::Callsite.capture_query_callsite(collection: self.collection_name, allow_disabling: true) do
+  for_each_this do |obj|
+  end
+end
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -67,7 +67,18 @@ pub fn format_node(ps: &mut ParserState, node: prism::Node) {
         Node::CallNode { .. } => {
             let call_node = node.as_call_node().unwrap();
             let is_last_call_in_chain = call_node.receiver().is_none();
-            format_call_node(ps, call_node, false, is_last_call_in_chain)
+
+            // Wrap standalone calls (no receiver) with blocks in a BreakableCallChainEntry
+            // so that line-length machinery can ignore nested blocks.
+            let has_block = call_node.block().and_then(|b| b.as_block_node()).is_some();
+
+            if is_last_call_in_chain && has_block {
+                ps.breakable_call_chain_of(MultilineHandling::Prism(false), |ps| {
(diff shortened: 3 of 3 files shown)


froydnj (approved):
Ahah, this explanation makes a lot of sense.

froydnj:
This removed ~90% of the changed lines when formatting Stripe's codebase with `--prism`.

The remaining cases appear to be, in no particular order:

1. indentation issues around call chains with arefs in them -- the tokens `[`, `index-expression`, `]` get split over multiple lines and ripper and prism disagree about how much the latter two get indented
2. same thing with block parameters that get split over multiple lines
3. similar indentation issues, but dealing with the last thing after an aref in a chain (only when the post-aref thing has a block?) -- these produce very noisy diffs because the block gets completely reindented
4. comment indentation changes, which I think are uniformly to prism's benefit
5. blank lines being deleted in some cases
6. blank lines being inserted in some cases
7. some weird comment shifting issues
8. a few heredoc-related things -- I can't immediately tell whether these are equivalent to the issues already filed
9. existing bugs already filed

I will try to file issues for some of the above today.

PR #714, opened 2025-12-23
[prism] Never use parens with `next`

### LINKED ISSUE #696: [prism] `next` with a multi-line argument introduces extra newlines and parens
Input, unchanged by ripper:

```ruby
Foo.each do |e|
  next Thing.new(
    x,
    y,
    z
  )
end
```

`rubyfmt --prism` gives:

```ruby
Foo.each do |e|
  next(
    Thing.new(
      x,
      y,
      z
    )
  )
end
```

My sense is that this is not a desirable change.

cc @reese

reese:
Closes #696

### DIFF
--- a/fixtures/small/next_with_args_expected.rb
+++ b/fixtures/small/next_with_args_expected.rb
@@ -1,3 +1,11 @@
 [].each do
   next 1
+  next Foo.new(
+    a,
+    b
+  )
+  next (Foo.new(
+    a,
+    b
+  ))
 end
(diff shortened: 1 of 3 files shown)


fixtures/small/next_with_args_expected.rb
```diff
@@ -1,3 +1,11 @@
 [].each do
   next 1
+  next Foo.new(
+    a,
+    b
+  )
+  next (Foo.new(
+    a,
+    b
+  ))
```
reese, line 10: This may seem kinda weird, but both Ripper and Prism report this as a `NextNode` with an argument which is a `ParenthesesNode`, so this is technically accurate -- it's not arg parens, it's a parens node.

froydnj (approved):

Issue #708, opened 2025-12-23
[prism] implicit ensure with block calls overindent the first line


froydnj:
I don't know a better way to describe this.

```ruby
module Test
  module Surgery
    describe "we ensure" do
      it "calls the function" do
        # call the function
        the_function.call
      ensure
        panic
      end
    end
  end
end
```

`rubyfmt --prism` produces:

```ruby
module Test
  module Surgery
    describe "we ensure" do
      it "calls the function" do
                # call the function
        the_function.call
      ensure
        panic

      end
    end
  end
end
```

The problem isn't strictly limited to comments, though:

```ruby
module Test
  module Surgery
    describe "we ensure" do
      it "calls the function" do
        ENV["CALL"] = "the function"
        the_function.call
      ensure
        ENV["CALL"] = nil
      end
    end
  end
end
```

Formatting with prism produces:

```ruby
module Test
  module Surgery
    describe "we ensure" do
      it "calls the function" do
                ENV["CALL"] = "the function"
        the_function.call
      ensure
        ENV["CALL"] = nil

      end
    end
  end
end
```

froydnj:
Oh, and notice the bonus blank line thrown in at the end.

PR #719, opened 2025-12-23
[prism] Properly wind to closing paren for lambda args

### LINKED ISSUE #716: [prism] extra newline in multilined lambda node returning a multilined hash
```ruby
entry = -> (
  stat,
  hive,
  bees,
  cap,
  version = nil,
  design = nil,
  effect = false
) {
  {
    "stat" => stat,
    "hive" => hive,
    "bees" => bees
  }
}
```

`rubyfmt --prism` adds an extra line at the start of the block:

```ruby
entry = -> (
  stat,
  hive,
  bees,
  cap,
  version = nil,
  design = nil,
  effect = false
) {

  {
    "stat" => stat,
    "hive" => hive,
    "bees" => bees
  }
}
```

reese:
Closes #716 

I assumed that `format_list_like_thing` would have done this line winding, but it only does so [if there's a comment there to capture](https://github.com/fables-tales/rubyfmt/blob/d2b0918eec2106a3477a5e55aad21b67a8810688/librubyfmt/src/parser_state.rs#L152-L153) on the following lines, so we also should go ahead and set the line here as well.

### DIFF
--- a/fixtures/small/lambda_actual.rb
+++ b/fixtures/small/lambda_actual.rb
@@ -11,3 +11,13 @@ lambda { |x| x }
 
   # But we're putting comments in it anyways!
 }
+
+-> (
+  foo,
+  bar
+) {
+  {
+    foo: foo,
+    bar: bar
+  }
+}
--- a/fixtures/small/lambda_expected.rb
+++ b/fixtures/small/lambda_expected.rb
@@ -12,3 +12,13 @@ lambda { |x| x }
 
   # But we're putting comments in it anyways!
 }
+
+-> (
+  foo,
+  bar
+) {
+  {
+    foo: foo,
+    bar: bar
+  }
+}
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -2669,6 +2669,7 @@ fn format_block_parameters_names(
             format_list_like_thing(ps, locals, end_offset, true);
         });
     }
+    ps.wind_dumping_comments_until_offset(end_offset);
 }
 
 fn format_block_local_variable_node(


froydnj (approved):
Love how adding functions dealing with comments fixes a testcase that has no comments.

reese:
Well the name of that method is _slightly_ misleading. It does wind forward for comments, but it also does one notable other thing, which is that it "winds" by going forward lines one-by-one, which circumvents the behavior of `at_offset` (or `on_line` in the Ripper code) where we add extra newlines if you call `at_offset` with a blank line in between. Generally at the end of delimeter-bound things like params, arrays, etc. we "wind" instead of just calling `at_offset` so we don't leave behind random blank lines.

The difference being that in theory you'd get different outputs between the two:

```ruby
# Original
-> (
  foo,
  bar

) { }

# Formatted with `at_offset` at the closing loc
-> (
  foo,
  bar

) { }

# Formatted with `wind_dumping_comments_until_offset` at closing loc
-> (
  foo,
  bar
) { }
```

So even though it mentions comments in the name, it winds forward for other reasons too.

froydnj:
> Generally at the end of delimeter-bound things like params, arrays, etc. we "wind" instead of just calling `at_offset` so we don't leave behind random blank lines.

Oooh, is it possible that this behavior is responsible for the random blank line in #708?

I think there might have been some other blank lines in the Stripe reformatting, I will have to check.

reese:
Yeah most likely, I haven't looked at that yet but it makes me think we're calling `at_offset` too early somewhere or possibly are missing one. (There's probably some other strange cause for the indentation issue though.)

PR #726, opened 2025-12-25
[prism] correctly handle implicit `BeginNode` as a wrapper node

### LINKED ISSUE #708: [prism] implicit ensure with block calls overindent the first line

froydnj:
Fixes #708 

With "implicit" begin nodes, the node really serves as a wrapper for its enclosed statements, so we should treat it similarly to a `StatementsNode` with respect to indentation and newline handling.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -15,11 +15,23 @@ pub fn format_node(ps: &mut ParserState, node: prism::Node) {
     use prism::Node;
 
     ps.at_offset(node.location().start_offset());
-    // StatementsNode is the only real "wrapper" node, meaning it purely contains
-    // other statements which themselves would be at the start of a line.
-    // We just ignore it here -- the alternative would be callers might need to have
-    // `ps.with_start_of_line(false, ...` for statements, which is semantically confusing
-    if ps.at_start_of_line() && !matches!(node, Node::StatementsNode { .. }) {
+
+    // StatementsNode is a "wrapper" node, meaning it purely contains other statements
+    // which themselves would be at the start of a line.  We just ignore it here -- the
+    // alternative would be callers might need to have `ps.with_start_of_line(false, ...`
+    // for statements, which is semantically confusing.
+    //
+    // BeginNode without a location for `begin` occurs in `def f; ... rescue; ... end` or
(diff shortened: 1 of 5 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -15,11 +15,23 @@ pub fn format_node(ps: &mut ParserState, node: prism::Node) {
...
+    // for statements, which is semantically confusing.
+    //
+    // BeginNode without a location for `begin` occurs in `def f; ... rescue; ... end` or
+    // similar and also -- at least for its `statements` field -- serves a similar
+    // wrapping function.
+    let needs_handling = ps.at_start_of_line()
+        && !(matches!(node, Node::StatementsNode { .. })
+            || node
+                .as_begin_node()
+                .map(|b| b.begin_keyword_loc().is_none())
```
reese, line 31: Might be worth unifying all the places we do this in one function or something, since we also have pretty similar cases handled in [various other places](https://github.com/fables-tales/rubyfmt/blob/95a0ea5e1af3e691cc68efffa57f02ef05f50956/librubyfmt/src/format_prism.rs#L2528-L2531), not sure if it's better to unify them in `format_node` or leave them handled in callers though 🤷

reese (approved):

PR #732, opened 2025-12-26
Use `Vec::splice` to insert tokens instead of `Vec::split_off`

reese:
`Vec::split_off` allocates a new intermediate Vec on every single token insertion (!), and then promptly deallocates it when appended to the new Vec. `splice` will instead move all the elements at the end (which in theory should generally be a single item, a newline token) to the right with a single `memmove` and then insert the `input` Vec in without an intermediary Vec.

I was looking at a flamegraph (see #731) and saw that we spent a significant amount of time each run allocating in `insert_at`, so this should cut down a decent amount of memory churn.


fables-tales (approved):

froydnj:
Nice!  I thought about doing this, but I wasn't sure if the splice implementation was actually going to be more efficient.

We could go further and accept an iterator in this function, and avoid the vec allocation for the argument entirely?

PR #733, opened 2025-12-26
Store comments in a `Vec` instead of a `BTreeMap`

reese:
A big chunk of formatting time is spent purely on inserting comments into a `BTreeMap`. Since it's a tree, it allocates for every single comment, and insertion into a BTreeMap is O(log n). I think the original intention was that a BTreeMap would be more efficient when checking if keys are present, but that's still O(log n), which is equivalent to doing a binary search on a Vec. For the most part, we also operate on comments in-order, so `.iter().any()` on a Vec is probably just as if not more efficient than the O(log n) membership check of a BTreeMap. On top of that, a Vec allocates less, is faster to insert, `drain` is more efficient than `split_off` on a tree, and a contiguous memory structure will hopefully have better cache locality than a tree structure.

My quick verification of this is that comment handling essentially disappears entirely from the flamegraph after this change.

### DIFF
--- a/librubyfmt/src/file_comments.rs
+++ b/librubyfmt/src/file_comments.rs
@@ -1,5 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, HashSet};
-use std::mem;
+use std::collections::{BTreeSet, HashSet};
 
 use crate::comment_block::CommentBlock;
 use crate::parser_state::line_difference_requires_newline;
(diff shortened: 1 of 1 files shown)


librubyfmt/src/file_comments.rs
```diff
@@ -180,7 +181,7 @@ impl FileComments {
                 sled.add_line(l);
             }
             _ => {
-                self.other_comments.insert(line_number, l);
+                self.other_comments.push((line_number, l));
```
froydnj, line 193: I _think_ this is the only place comments get added to this field -- WDYT about adding a debug assertion here that either this vec is empty or the line number that we're pushing is greater than the last line number in the vector?  Just as a way to make sure this is sorted, since the bits below depend on that invariant.

librubyfmt/src/file_comments.rs
```diff
@@ -63,7 +62,7 @@ impl LineIndex {
 #[derive(Clone, Debug, Default)]
 pub struct FileComments {
     start_of_file_contiguous_comment_lines: Option<CommentBlock>,
-    other_comments: BTreeMap<LineNumber, String>,
+    other_comments: Vec<(LineNumber, String)>,
```
froydnj, line 66: A comment here about sortedness would be a good idea.

froydnj (approved):

PR #734, opened 2025-12-26
Use a shared `&'static str` for indentation

reese:
Currently we're allocating a new `String` for every single indent. This is silly -- indentation is essentially the same handful of values every time: 2 spaces, 4 spaces, 6 spaces, etc. What I've done in this PR is generate a large `&str` at compile time that's 128 spaces long (it falls back to allocating if you go over 128, but also if you're over 64 blocks deep in a single Ruby file then I would be getting worried) that we just read string slices from, which means we should ~never have to allocate anything for indentation.

### DIFF
--- a/librubyfmt/src/util.rs
+++ b/librubyfmt/src/util.rs
@@ -1,4 +1,21 @@
+use std::borrow::Cow;
+
 use ruby_prism::{ConstantId, Location};
+
+/// Maximum indent depth we cache without allocation.
+const MAX_CACHED_INDENT: usize = 128;
+/// A pre-allocated string of spaces for efficient indentation
+const SPACES: &str = unsafe { std::str::from_utf8_unchecked(&[b' '; MAX_CACHED_INDENT]) };
+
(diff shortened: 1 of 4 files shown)


librubyfmt/src/util.rs
```diff
@@ -1,5 +1,22 @@
+use std::borrow::Cow;
+
 use ruby_prism::{ConstantId, Location};
 
+/// Maximum indent depth we cache without allocation.
+const MAX_CACHED_INDENT: usize = 128;
+/// A pre-allocated string of spaces for efficient indentation
+const SPACES: &str = unsafe { std::str::from_utf8_unchecked(&[b' '; MAX_CACHED_INDENT]) };
```
reese, line 8: We could just have a giant string of 128 spaces, but that felt weird and also seemed potentially error-prone (how would any reviewer know I put the correct number of spaces in? what if we changed it? etc.) so I opted to just generate it at compile-time.

froydnj (approved):
Yes!

PR #735, opened 2025-12-26
Use iterators for `insert_at`

reese:
This was a [suggestion](https://github.com/fables-tales/rubyfmt/pull/732#issuecomment-3693345328) by @froydnj to take #732 one step further: instead of taking a Vec, almost all of the callers already have iterators, so instead of `collect`ing into Vecs and then moving them back into iterators with `drain`, we can just keep the iterators and pass them directly. This should keep us from allocating a few more intermediary Vecs.

### DIFF
--- a/librubyfmt/src/comment_block.rs
+++ b/librubyfmt/src/comment_block.rs
@@ -23,16 +23,13 @@ impl CommentBlock {
         self.comments.push(line);
     }
 
-    pub fn into_line_tokens(self) -> Vec<ConcreteLineToken> {
-        self.comments
-            .into_iter()
-            .flat_map(|c| {
-                vec![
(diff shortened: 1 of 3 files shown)


froydnj (approved):
I think I would genericize the insertion functions (and everything up the call chain) over their iterator argument type so you a) don't have to box things up -- which avoids allocations -- and b) you don't have to do dynamic dispatch down the line, which would be slower.

reese:
Yeah I'm going to do that in a follow-on PR -- it's mechanical but a bit invasive since you basically have to follow the callers all the way up to `breakable_entry_stack`, but it will get rid of pretty much all the remaining Boxing and dynamic dispatch

PR #736, opened 2025-12-26
 Avoid cloning and converting tokens when doing line length checks 

reese:
For breakables, we currently check if it will go over the 120 character line limit by rendering the tokens into a string and then looking at the length of the string. We don't actually need to do this -- we know how long the tokens will be based on their contents, so we could just look at that and have tokens report their (future) length, allowing us to avoid cloning the whole token array, doing string conversions, and so on.

The most notable refactor is in how call chain breakables work. There's a lot of token munging happening in there, the main one requiring mutation being the way we handle brace blocks. For the last brace block in a chain, we ignore the contents of the block and only look at the params. Since we want to avoid mutations, we instead have the token do that ignoring in a helper method.

### DIFF
--- a/librubyfmt/src/render_targets.rs
+++ b/librubyfmt/src/render_targets.rs
@@ -283,14 +283,14 @@ impl AbstractTokenTarget for BreakableCallChainEntry {
                 // Pop off all tokens that make up the block (but not the block params!),
                 // since we assume that the block contents will handle their own line
                 // length appropriately.
-                while let Some(token) = tokens.last() {
+                while let Some((token, rest)) = tokens.split_last() {
                     if matches!(
                         token,
                         AbstractLineToken::BreakableEntry(BreakableEntry { delims, .. }) if *delims == BreakableDelims::for_block_params()
                     ) {
                         break;
                     }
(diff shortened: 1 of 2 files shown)


librubyfmt/src/render_targets.rs
```diff
@@ -275,37 +290,25 @@ impl AbstractTokenTarget for BreakableCallChainEntry {
...
+                brace_block_params_only_index = Some(index);
             }
         }
 
         if let Some(AbstractLineToken::BreakableEntry(_)) = tokens.first() {
-            tokens.remove(0);
+            if let Some(idx) = brace_block_params_only_index {
+                brace_block_params_only_index = Some(idx - 1);
+            }
+            tokens = &tokens[1..];
```
froydnj: FWIW, there's `split_{first,last}` methods on slices that might be a tad more efficient that checking for a first element and then re-slicing.  Here and below.

froydnj (approved):
I tried to make the slice bits work for `BreakableCallChainEntry`, but ran into some issues about modifying the tokens, and wasn't smart enough to see that we needed a new method.  I'm glad you were able to make it work!

PR #740, opened 2025-12-28
fix paren formatting

fables-tales:


reese:
As far as the approach here, I think probably what you want is basically reverting the changes in 764836d5119596fdd89e875bb9edaf9a3c526c77 and then changing [this bit](https://github.com/fables-tales/rubyfmt/commit/764836d5119596fdd89e875bb9edaf9a3c526c77#diff-323c3a8194f72a3b48bc6bacd480710193b1bb4a86bc56a4cbb891c87e9e5523L4413-L4416) to have more restrictions -- i.e. return `None` if there's multiple statements _or_ it's one of whatever list of node types (`UnlessNode`, `IfNode`, `UntilNode`, etc.). (I worry that there will be ~a lot of node types that you'll have to exempt -- and you'll probably want to add a bunch of tests for return/yield/every other keyword where we sometimes don't render them with parens --  hence my general FUD with stripping parens, but c'est la vie.)

(Gonna unassign myself while you're working on this, but feel to reassign when things are working!)

Issue #713, opened 2025-12-23
[prism] extra indentation for array index expressions in the middle of call chains


froydnj:
```ruby
assert_equal(
  1,
  T.cast(T.must(pages[1]&.elements)[0], Name::Spacing::To::Extend::CodeQuality::AcrossManyLines).table[
    :line_items
  ]
    .length
)
```

is unchanged by ripper, but `rubyfmt --prism` gives:

```ruby
assert_equal(
  1,
  T.cast(T.must(pages[1]&.elements)[0], Name::Spacing::To::Extend::CodeQuality::AcrossManyLines).table[
      :line_items
    ]
    .length
)
```

which...I think I like better?  I don't love how many changes this induces on pay-server, but I think it is more consistent with call chain formatting generally.

Curious about other opinions here.

PR #743, opened 2026-01-01
Upgrade jemalloc crate

reese:
On the tin.

The `jemalloc` crate itself has not been maintained for many years and was later forked by the TiKV folks, so this switches over to that. The [changelog](https://github.com/tikv/jemallocator/blob/main/CHANGELOG.md) doesn't seem to have many major changes other than keeping up with some newer jemalloc versions and removing a deprecated `fs_extra` dep.

(Long term, I'm curious if we still want/need to use a custom allocator for such short-lived processes, but I'll leave that for a separate discussion/benchmarking/etc.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,12 +416,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
-name = "fs_extra"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2022715d62ab30faffd124d40b76f4134a550a87792276512b18d63272333394"
-
-[[package]]
 name = "getrandom"
 version = "0.3.2"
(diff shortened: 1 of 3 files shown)


froydnj (approved):
My suspicion is that jemalloc matters a lot more for ripper, since jemalloc is backing an entire Ruby VM.

OTOH, I remember doing benchmarking on parsing performance on Stripe's codebase when Prism was in development, and it was definitely noticeable when I forgot to set the environment variables that switched on jemalloc.

reese:
> when Prism was in development

OOC, does this imply that you're live with the prism implementation now? 👀 (Wasn't sure if maybe #713 was blocking or something)

froydnj:
> > when Prism was in development
> 
> OOC, does this imply that you're live with the prism implementation now? 👀 (Wasn't sure if maybe #713 was blocking or something)

No, this was referring to the actual development of Prism itself; I landed a decent number of parsing performance improvements.

FWIW, my current TODO for the prism upgrade is to get the rubyfmt bumps in that change ripper's formatting along with reformatting the codebase, and then we'll see where prism incompatibilities stand (modulo the open issues where the prism behavior is an improvement).

PR #745, opened 2026-01-01
[prism] Avoid allocating Strings by threading `'src` lifetime throughout parser state and node types

reese:
The biggest remaining source of unnecessary String allocations is simple constants like identifiers that come from source code. Since historically we haven't operated by references (since we had to copy all the Ripper objects) back to source nodes, most of the main structs (`ParserState` etc.) don't have clear lifetimes, forcing most of our render tokens to operate entirely on owned Strings.

This PR is a bit invasive but ultimately relatively straightforward: it adds a `'src` lifetime, adds it throughout all of the structs and functions that operate on Prism nodes or tokens, and then updates places to use `{loc,const}_to_str` wherever possible. This means we no longer need to clone identifiers or convert them to strings, including in some of the associated machinery (e.g. variable bindings).

It's probably easiest to review this commit-by-commit. The first commit mostly adds lifetimes to structs, the second to formatting functions, and the last converts `*_to_string` to `*_to_str` wherever possible. It's possible this same thing could be applied to other callers (Strings, heredoc closers, etc.) but for the sake of minimizing the change I've punted on those for later.

### DIFF
--- a/librubyfmt/src/comment_block.rs
+++ b/librubyfmt/src/comment_block.rs
@@ -24,7 +24,7 @@ impl CommentBlock {
         self.comments.push(line);
     }
 
-    pub fn into_line_tokens(self) -> impl Iterator<Item = ConcreteLineToken> {
+    pub fn into_line_tokens<'src>(self) -> impl Iterator<Item = ConcreteLineToken<'src>> {
         self.comments.into_iter().flat_map(|c| {
             [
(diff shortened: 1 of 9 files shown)


froydnj (approved):
Oh man, I thought this was going to have to wait until the ripper backend was gone.  So nice.

PR #756, opened 2026-01-04
Fix multilining of call writes

### LINKED ISSUE #752: [prism] refusing to line-break extremely long csend call chains
Input:

```ruby
module Test
  class TestThing
    describe "process" do
      describe "inspect" do
        it "does the right thing" do
          v2_request
            .additional_data_private
            &.sub_elements
            &.practical_mechanics_signs_etcs
            &.practical_safety_indications_&.specific_collect_pointer = Constant::Thing
        end
      end
    end
  end
end
```

`rubyfmt --prism` turns this into one very long line:

```ruby
module Test
  class TestThing
    describe "process" do
      describe "inspect" do
        it "does the right thing" do
          v2_request.additional_data_private&.sub_elements&.practical_mechanics_signs_etcs&.practical_safety_indications_&.specific_collect_pointer = Constant::Thing
        end
      end
    end
  end
end
```

I think there is something wrt the method names that triggers different behavior in ripper and prism.  For instance,

```ruby
module Test
  class TestThing
    describe "process" do
      describe "inspect" do
        it "does the right thing" do
          v2_request.additional_data_private&.sub_elements&.pratical_mechanics_signs&.practical_safety_indicators&.specific_collection_pointer = Constant::Thing
        end
      end
    end
  end
end
```

is formatted unchanged by both, despite being over the line length both in the overall expression and the LHS.  (I don't think the latter is desirable, but it is what it is today.)

reese:
Closes #752

This is a two-for-one PR that fixes two different bugs for calls where the left-hand side is a call chain that should break to multiple lines.

For Prism, attr_write call chains were always single-lined, I believe largely as a workaround for the left and right sides being wrapped in the same breakable. This separately formats the left- and right-hand side in their own breakables so that they can multiline independently of each other.

For Ripper, the issue was that `Field` nodes were not handled as call chains and had their own formatting logic that doesn't understand multilining. This now converts them into `CallChainElement`s so that they can use the regular call chain machinery.

Each change is done in its own commit.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -86,10 +86,10 @@ pub fn format_node<'src>(ps: &mut ParserState<'src>, node: prism::Node<'src>) {
 
             if is_last_call_in_chain && has_block {
                 ps.breakable_call_chain_of(MultilineHandling::Prism(false), |ps| {
-                    format_call_node(ps, call_node, false, is_last_call_in_chain);
+                    format_call_node(ps, call_node, false, is_last_call_in_chain, false);
                 });
             } else {
-                format_call_node(ps, call_node, false, is_last_call_in_chain)
+                format_call_node(ps, call_node, false, is_last_call_in_chain, false)
             }
         }
         Node::CallOperatorWriteNode { .. } => {
@@ -1711,6 +1711,7 @@ fn format_call_node<'src>(
     call_node: prism::CallNode<'src>,
     skip_receiver: bool,
     is_final_call_in_chain: bool,
+    skip_attr_write_value: bool,
 ) {
     let method_name = const_to_str(call_node.name());
     let end_offset = call_node.location().end_offset();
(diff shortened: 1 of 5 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2108,41 +2111,66 @@ fn format_call_chain<'src>(ps: &mut ParserState<'src>, call_node: ruby_prism::Ca
...
-    is_attr_write: bool,
 ) {
     if let Some(current) = segments.pop() {
         let has_inner = !segments.is_empty();
 
         let (chain_elements, trailing_arefs) = extract_trailing_arefs(current);
 
+        // Render the right-hand side of the write in its own breakable, so that each side can
+        // break independently of each other
+        let trailing_attr_write_value = chain_elements.last().and_then(|last| {
```
froydnj, line 2129: It would be nice if we could only do this calculation once, rather than for every node in the call chain...or are there cases where we could have multiple attribute writes in the call chain?  I guess the same thing goes for computing trailing arefs, too?

Those changes could be handled separately, if at all.

froydnj (approved):
FWIW, I think there are a few heredoc formatting issues that I need to reduce down, and a couple other miscellaneous things, but the bulk of the remaining reformatting on Stripe's codebase is #713, and we've decided we're going to live with that.

librubyfmt/src/format_prism.rs
reese: I think in practice this almost always does happen only once for the chain unless you have arefs in the middle of the chain, which is not the _most_ common construct. This runs once per "segment" where a segment is more-or-less call chains separated by arefs. That is, something `foo.bar[1].baz` would be two segments, `foo.bar[1]` and then `<previous-segment>.baz`. And for attr_writes specifically, I believe they'll always be exactly one segment so this only runs once.

(I grant that this segment stuff is slightly awkward, but it's because Prism handles arefs as calls instead of as their own separate construct, so we had to manually split up the chains into these segments to mirror the Ripper behavior.)

PR #757, opened 2026-01-05
[prism] Fix splitting logic for aref grouping

### LINKED ISSUE #713: [prism] extra indentation for array index expressions in the middle of call chains

reese:
Closes #713 

I kinda goofed the logic on grouping arefs as part of a segment when I first wrote this. We shouldn't split before an aref -- we should instead split _after_ that aref, and we also would want to split after _all_ the arefs in that group. This was causing all the "trailing aref" logic to not work quite as expected so arefs were still getting rendered inside a breakable (hence the extra indentation).

It was unclear to me if we preferred the indentation of the Prism or Ripper implementations, but IMO we should shoot for the same output first and then we can adjust the formatting later if we want -- I'd rather that be an intentional choice and not an accident. (Even then, I think this is more correct and would make an intentional choice to further indent it ~trivial.)

### DIFF
--- a/fixtures/small/aref_in_call_chain_actual.rb
+++ b/fixtures/small/aref_in_call_chain_actual.rb
@@ -1,0 +1,29 @@
+assert_equal(
+  1,
+  T.cast(T.must(pages[1]&.elements)[0], Name::Spacing::To::Extend::CodeQuality::AcrossManyLines).table[
+    :line_items
+  ]
+    .length
+)
+
+foo.bar[some_really_long_index][another_really_long_index]
+  .baz
+  .qux
+
+result.data[
+  :first_key,
+  :second_key
+]
+  .transform
+  .process
+
+outer_method(
+  inner.call[0]
+    .chain
+    .another
+)
+
+matrix[row][col]
+  .value
+  .to_s
+  .upcase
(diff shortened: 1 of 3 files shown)


froydnj (approved):
I haven't checked this on Stripe's codebase, but the tests look like they do the right thing.

froydnj:
And FWIW, the recent spate of performance improvements (0.11.211 - > 0.11.233) seem to improve whole-codebase formatting with ripper by ~25% on Stripe's codebase 🎉 .

PR #758, opened 2026-01-05
[prism] Don't panic on invalid UTF-8 strings in heredocs

reese:
The way we determine the "common indent" for a heredoc (that is, for squiggly heredocs, the farthest-left line that determines the indentation level) is to look for lines where the `unescaped` and the content_loc have different leading whitespaces. We previously did this by converting them to strings and looking at their length, but this panics on invalid UTF-8 because all the `*_to_str` functions assume they are given valid UTF-8 strings.

Instead, this PR just counts the whitespace chars directly at the beginning of each slice, which should achieve the same thing without panicking on invalid UTF-8.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -8,7 +8,7 @@ use crate::{
     parser_state::{FormattingContext, HashType, ParserState},
     render_targets::MultilineHandling,
     types::SourceOffset,
-    util::{const_to_str, loc_to_str, loc_to_string, u8_to_str},
+    util::{const_to_str, loc_to_str, loc_to_string},
 };
 
 pub fn format_node<'src>(ps: &mut ParserState<'src>, node: prism::Node<'src>) {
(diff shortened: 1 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -909,12 +909,14 @@ fn format_inner_string<'src>(
...
+                    let raw = node.content_loc().as_slice();
+                    let unescaped = node.unescaped();
 
-                    // Count leading whitespace in each
-                    let raw_leading = raw.len() - raw.trim_start().len();
-                    let unescaped_leading = unescaped.len() - unescaped.trim_start().len();
+                    let raw_leading = raw.iter().take_while(|&&b| b == b' ' || b == b'\t').count();
+                    let unescaped_leading = unescaped
+                        .iter()
+                        .take_while(|&&b| b == b' ' || b == b'\t')
```
froydnj, line 918: I know that Sorbet's heredoc bits in the parser have to take into account a tab being 8 spaces.  Do we have to do that same accounting here as well?

Answering my own question, I guess not, because the number of spaces and tabs in the raw and unescaped are essentially coming from the same source, so things should just match up?

froydnj (approved):
Thanks for this!  My efforts over the weekend to format individual files to try to find the invalid UTF8 were weirdly _not_ crashing (vs. the format-in-one-go invocation that was), and I hadn't yet debugged what I did wrong.  Ideally this will just magically fix things.

librubyfmt/src/format_prism.rs
reese: Yeah I _think_ in this case since they're the same source, it doesn't really matter, and I think if you had a tab that was beyond the common indent, it would be a part of the string contents and just get rendered anyways.

PR #765, opened 2026-01-06
[prism] Fix extra newline after require kwarg

### LINKED ISSUE #762: [prism] `require:` kwarg causes blank lines to be inserted
Input:

```ruby
ThingToCall.call(
  require: "not a regular require"
)
```

`rubyfmt --prism` produces:

```ruby
ThingToCall.call(
  require: "not a regular require"

)
```

I think something about how prism is emitting kwarg tokens is causing line metadata to get its `require` bit set, which messes everything up.  But I don't obviously see the differentiation between how ripper does it and how prism does it.

reese:
Closes #762 

[This bit](https://github.com/fables-tales/rubyfmt/blob/bb8c4a260db719b57f038868d57e6753feaf7a53/librubyfmt/src/intermediary.rs#L133-L137) in the line metadata handling looks for `"require"` as a DirectPart to determine when we're in a series of `require` calls, which we insert a blank line after. This was happening after `require:` as a kwarg (or hash key etc.) in the Prism implementation because in Ripper, keys would always be a DirectPart of `"name:"`, whereas Prism separately emitted the name and then a colon as two parts.

This PR tries to always emit assoc node keys as `"name:"` to circumvent this. I don't know if it's my favorite approach in the world (part of me wonders if we should instead have more descriptive token types that aren't just `DirectPart` for this), but this is what Ripper does, so I figured I'd keep it simple and just follow that.

### DIFF
--- a/fixtures/small/other_requires_actual.rb
+++ b/fixtures/small/other_requires_actual.rb
@@ -3,6 +3,7 @@ ThingToCall.call(
   other: "value"
 )
 
+require = bees!
 foo(require: bar)
 
 {
--- a/fixtures/small/other_requires_expected.rb
+++ b/fixtures/small/other_requires_expected.rb
@@ -3,6 +3,7 @@ ThingToCall.call(
   other: "value"
 )
 
+require = bees!
 foo(require: bar)
 
 {
--- a/fixtures/small/require_kwarg_actual.rb
+++ b/fixtures/small/require_kwarg_actual.rb
@@ -1,0 +1,20 @@
+ThingToCall.call(
+  require: "not a regular require",
+  other: "value"
+)
+
+foo(require: bar)
+
+{
+  :require => foo,
+  :other => 1
+}
+
+"require#{foo}"
+
+x = {require: 1, other: 2}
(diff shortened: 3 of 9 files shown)


froydnj (changes requested):
idk if I'm a fan of this approach -- I think there's a lot of complexity around symbol nodes (e.g. `'require':`) that I'm not sure this is going to handle.  If nothing else, I think there should be some more tests around hashes with `:require` symbol keys in various formats (single-quoted, double-quoted, etc.).  I know this same issue comes up with `require:` keys in hashes, so it'd be good to have tests for those regardless.

I do think having a proper `MethodName` token or some such that we can check for `"require"` instead of the current approach would be better and more robust overall.  That would also handle the issue where we format `require = foo` with newlines after it, even though there isn't anything to require that.

froydnj (approved):
Nice, thank you!

froydnj:
This seems to have introduced a very few cases in ripper where a heredoc immediately followed by a receiver-less method call overindents the method call on Stripe's codebase.  Trying to minimize a test case.

PR #766, opened 2026-01-06
[prism] Match Ripper's multilining logic for call chains beginning with literal expressions

### LINKED ISSUE #764: [prism] less indentation on array/hash literal calls with blocks
Eyeballing things, this seems to be the biggest source of formatting differences between ripper and prism, at least in terms of sheer lines changed.  Input, unchanged by ripper:

```ruby
module Intake
  class Category
    def self.get(name)
      if name == :fields
        core_fields = [
          CommonFields::OBVIOUSLY,
          CommonFields::THESE,
          CommonFields::ARE,
          CommonFields::FAKE_RESOURCE.with_wombo(true).with_combo(true).with_explosion(true)
        ]
          .map do |field|
            # Swap things out, to make sure.
            override = self.override_fields.find { |override_field| override_field.slug == field.slug }
            override || field
          end

        core_fields + self.additional_fields
      else
        super
      end
    end
  end
end
```

`rubyfmt --prism` turns this into:

```ruby
module Intake
  class Category
    def self.get(name)
      if name == :fields
        core_fields = [
          CommonFields::OBVIOUSLY,
          CommonFields::THESE,
          CommonFields::ARE,
          CommonFields::FAKE_RESOURCE.with_wombo(true).with_combo(true).with_explosion(true)
        ].map do |field|
          # Swap things out, to make sure.
          override = self.override_fields.find { |override_field| override_field.slug == field.slug }
          override || field
        end

        core_fields + self.additional_fields
      else
        super
      end
    end
  end
end
```

so we don't break the `.map` call onto a separate line, which changes the indentation of the block.  This only seems to happen with array (and maybe sometimes with hash?) literals in the receiver position.

reese:
Closes #764 

For call chains, we do some ~finicky logic for cases where the first item in a call chain is an expression like an array/hash literal. This is to accommodate some common use cases where something like calling `.freeze` or `.map(&:foo)` after a big multiline expression. However, if there's comments in the middle of all this, we'll still need to fully multiline it, but the Ripper implementation multilines [if there's comments _anywhere_ in call chain's tokens](https://github.com/fables-tales/rubyfmt/blob/trunk/librubyfmt/src/render_targets.rs#L365-L370), whereas the Prism one only was looking if they're [between the end of the first expression and the first dot operator](https://github.com/fables-tales/rubyfmt/blob/02fb1545310998c22a27b28833c3d3b90dbf97eb/librubyfmt/src/format_prism.rs#L2341-L2344).

(For the record, I think the Prism logic is probably more correct, but I'm fine with shooting for as much consistency with Ripper as possible for now, and we can explicitly choose this later.)

### DIFF
--- a/fixtures/small/array_call_chain_with_block_comment_actual.rb
+++ b/fixtures/small/array_call_chain_with_block_comment_actual.rb
@@ -1,0 +1,10 @@
+[
+  CommonFields::OBVIOUSLY,
+  CommonFields::THESE,
+  CommonFields::ARE,
+  CommonFields::FAKE_RESOURCE.with_wombo(true).with_combo(true).with_explosion(true)
+]
+  .map do |field|
+  # Swap things out, to make sure.
+  override || field
+end
--- a/fixtures/small/array_call_chain_with_block_comment_expected.rb
+++ b/fixtures/small/array_call_chain_with_block_comment_expected.rb
@@ -1,0 +1,10 @@
+[
+  CommonFields::OBVIOUSLY,
+  CommonFields::THESE,
+  CommonFields::ARE,
+  CommonFields::FAKE_RESOURCE.with_wombo(true).with_combo(true).with_explosion(true)
+]
+  .map do |field|
+    # Swap things out, to make sure.
+    override || field
+  end
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -2340,7 +2340,7 @@ fn call_chain_elements_are_user_multilined<'src>(
         // placement, the user probably intended to break this onto multiple lines anyways.
         let has_comment = ps.has_comment_in_offset_span(
             call_chain_elements[0].location().end_offset(),
-            start_loc_for_call_node_in_chain(&call_chain_elements[1].as_call_node().unwrap()),
+            call_chain_elements.last().unwrap().location().end_offset(),
         );
         if is_literal_expression && !has_comment {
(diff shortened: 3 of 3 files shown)


froydnj (approved):
Don't mind me, I'm just completely misdiagnosing the issues here. 🤦‍♂️ 

Thanks for this fix!

froydnj:
🤦‍♂️ Of course, testing this out, there are now (a lot more?) cases where we aggressively multiline, so I think this netted out to more changes overall.

reese:
Yeah I mean that's not wildly surprising -- I think checking for comments anywhere in the chain is probably ~not quite right, especially since in the Ripper implementation this was looking for _tokens_, and it didn't recurse so it probably isn't including comments inside breakables (like comments on method params in the chain, etc.). I'm not 100% sure how we could reliably reproduce that with Prism, the token check is pretty fundamentally different.

Depending on how many there are and how feasible/annoying it is to land changes and so on, personally I think the Prism implementation before this change was the correct one, and I'd be fine to just revert it with the understanding that it would lead to changes.

reese:
@froydnj Were you thinking we revert this or did you want approach it another way? Personally I'm fine to revert (and like I said, I think the original was more correct anyways)

froydnj:
> @froydnj Were you thinking we revert this or did you want approach it another way? Personally I'm fine to revert (and like I said, I think the original was more correct anyways)

I was thinking reverting was the right thing, I just didn't get to it yesterday.  Was planning to do so today unless you want to beat me to it.

PR #770, opened 2026-01-07
Revert "[prism] Match Ripper's multilining logic for call chains beginning with literal expressions"

reese:
Reverts fables-tales/rubyfmt#766


froydnj (approved):
This is fine by me.  I wish we had more testcases around this stuff.

idk if you have a post-prism checklist, but perhaps this testcase, slightly altered, would be a good candidate for committing after ripper disappears.

reese:
We can just leave it around as a Prism-specific test if that's worthwhile, I can revert the old change but leave parser-specific outputs

reese:
(Actually I'll do this in a separate PR, I'm gonna try and make a few test cases for this and try to dial in where they differ)

froydnj:
Oh, that's right, I forgot about the denylists for tests.  That'd be great!

PR #775, opened 2026-01-10
[prism] add a testcase for correctly binding masgns

froydnj:
I noticed that we get this correct in prism, but wrong in ripper.  I'm not really motivated to try to tease apart the intricacies here in ripper, but it would be good to have test coverage for this case.

### DIFF
--- a/fixtures/small/multi_variable_binding_expected.rb
+++ b/fixtures/small/multi_variable_binding_expected.rb
@@ -1,0 +1,9 @@
+def foo
+  some, method = self.method()
+  some, method = self.field.method
+end
+
+def bar
+  (other, method) = self.method()
+  (other, method) = self.field.method
+end
--- a/fixtures/small/multi_variable_binding_actual.rb
+++ b/fixtures/small/multi_variable_binding_actual.rb
@@ -1,0 +1,9 @@
+def foo
+  some, method = self.method()
+  some, method = self.field.method
+end
+
+def bar
+  (other, method) = self.method()
+  (other, method) = self.field.method
+end
--- a/tests/fixtures_test.rs
+++ b/tests/fixtures_test.rs
@@ -15,6 +15,7 @@ const DISABLED_RIPPER_TESTS: &[&'static str] = &[
     // https://github.com/fables-tales/rubyfmt/issues/474
     "small_args_forwarding_additional_args",
     "small_alias_global_var",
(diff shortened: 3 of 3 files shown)


reese (approved):
(FWIW, tests in `fixtures/{ripper,prism}` will run only for their respective implementation if you'd rather do that instead of adding more stuff to this denylist, up to you)

fixtures/small/multi_variable_binding_expected.rb
```diff
@@ -0,0 +1,9 @@
+def foo
+  some, method = self.method()
```
fables-tales, line 2: we should pick a style here, similar to how in other parts of rubyfmt we have a single possible output for multiple constructs. (i.e. we should pick either parentheses or not parentheses)

fables-tales (changes requested):

fixtures/small/multi_variable_binding_expected.rb
froydnj: The rule that we have is that `self.method` is parenthesized if `method` is a local variable in the current scope -- I think there are also limits on call chain length, so e.g. `self.method.other` is not parenthesized even if `method` and `other` are local variables.  You can see the testcases we have for this in  

https://github.com/fables-tales/rubyfmt/blob/861d89e100ec1caaaf4cbd5bcc53598238113561/fixtures/small/variable_binding_expected.rb

The ripper implementation doesn't correctly bind variables for masgns, but the prism implementation does.  So it would have been nice to put this in with the other `variable_binding` testcases, but until prism is the One True Way, we'll have to just note the difference and move on.

fables-tales: I meant the left hand side (i.e. `(some,method) =` vs `some, method =`

froydnj: > I meant the left hand side (i.e. `(some,method) =` vs `some, method =`

Ah, OK.  I'm down for doing that, but I think that's a separate PR, and shouldn't hold up this one.  WDYT?  (I also don't know if there are weird cases lurking where we really would need the parens.)

fables-tales: sure we can merge this but let's take an issue to do the other thing

fables-tales (approved):

PR #778, opened 2026-01-10
Model inline conditionals as breakables

reese:
Currently for mod conditionals (like `return if foo`), we transform those to multiple lines (`if foo \n return \n end`) if they're multiline/too long. The current implementation does this in a wildly inefficient manner: cloning the ~entire parser state, rendering the whole thing as a string, and then checking if the string is multiline. It does all this every time a mod conditional comes up, so it currently takes up a pretty decent chunk of profiling, not to mention it's kinda a hack.

This PR makes conditionals work like any other breakable. It stashes the tokens for the predicate and the body of the conditional, then uses those to render into either an inline or block form depending on the line length. This should more accurate and efficient (with _much_ less cloning/allocating) than rendering every single time.


reese:
Closes #829

fixtures/small/long_modifier_conditional_actual.rb
```diff
@@ -0,0 +1,5 @@
+some_really_long_method_name_that_takes_up_space! if another_really_long_condition_that_makes_this_line_exceed_one_hundred_twenty_chars
+
+some_really_long_method_name_that_takes_up_space! unless another_really_long_condition_that_makes_this_line_exceed_one_hundred_twenty
+
+short_method_call if a_condition_that_fits_within_the_line_length_limit_of_one_hundred_twenty
```
froydnj, line 5: Can you add testcases like the ones from #829?  idk if those are valuable considering that the fixes for #813 are kind of similar, but obviously #813 got fixed without addressing #829, so I think more testcase coverage is better.

librubyfmt/src/format_prism.rs
```diff
@@ -3757,47 +3757,29 @@ fn format_conditional_node<'src>(
...
-            Conditional::While(_) | Conditional::Until(_) => false,
+        match conditional {
+            Conditional::While(_) | Conditional::Until(_) => {
+                // Always use inline format for while/until modifiers
+                format_inline_conditional(
+                    ps,
+                    conditional.predicate(),
+                    conditional.statements(),
+                    conditional_keyword,
+                );
```
froydnj, line 3768: I don't know that it needs to be fixed here, but I do wonder whether there is something lurking with non-idempotent formatting by forcing this to always be on a single line.

froydnj (changes requested):
I think the testcases mentioned below would be good unless you feel strongly in the other direction.

librubyfmt/src/format_prism.rs
reese: Just for clarification, this is [equivalent to the existing behavior](https://github.com/fables-tales/rubyfmt/blob/544151464e60d6ca858b86c0e1044894ed887a22/librubyfmt/src/format_prism.rs#L3757-L3761), so at least as far as this change is concerned this is a no-op for `while`/`until`.

(Happy to follow up separately if there's a known issue with them, but those transformations are unsafe since inline/block while and until have different runtime behavior, so we never transform them. I think since we _never_ multiline them there shouldn't be latent issues around idempotency AFAICT, but open to being proven wrong there.)

froydnj: Yeah, I can see this change a) as a no-op for `while`/`until` and b) due to _never_ multilining them, we should be OK, but c) that still kind of makes me nervous.  Fortunately they don't seem to be widely used.

froydnj (approved):
Thanks for adding the testcase!

PR #782, opened 2026-01-11
Clean up misleading `is_comment` method

reese:
From ages long past (as in, it was written several years ago and never touched), there's a method called `is_comment` on `ConcreteLineToken` that for whatever reason checks for an `Indent` token. There's a method directly next to it called `is_indent` with the same implementation, and changing `is_comment` to actually check for comments breaks ~everything, so this PR just deletes it an replaces its few callers with `is_indent`.

(_Do as I do, not as I say_ etc.)


froydnj (approved):
🙈

PR #783, opened 2026-01-13
[prism] Bump ruby-prism to 1.8.0

reese:
On the tin. The main impetus for bumping besides just staying up to date is pulling in @froydnj 's [PR](https://github.com/ruby/prism/pull/3822) to add `first()` and `last()` to node lists. (I updated call sites to use those in the second commit!)


froydnj (approved):
🎉

Issue #739, opened 2025-12-28
paren expressions


fables-tales:
bl;uf: I feel like we should elide redundant parens when we can. e.g.:

```
if (true)
end
```

should become

```
if true
end
```

but on trunk this doesn't work:

```
➜  rubyfmt git:(trunk) ✗ echo "if (true); end" | cargo run --release -- --prism
    Finished `release` profile [optimized + debuginfo] target(s) in 0.13s
warning: the following packages contain code that will be rejected by a future version of Rust: fs_extra v1.2.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 3`
     Running `target/release/rubyfmt-main --prism`
if (true)
end
```

I understand for some types of expressions (e.g. if (1 unless rand) ) we need to keep the parens, however I feel very strongly that we should eliminate redundant parens. Both rustfmt and and gofmt have this behaviour.

reese:
BLUF: I'm not gonna block anything here, but I do think the `gofmt` and `rustfmt` comparisons aren't totally valid -- they're entirely different syntaxes! I'd love to have a documented reason we have this that's not just "rustfmt does this."

---

It's not really a hill I'm willing to die on, but I thought I'd fully explain something I mentioned in Discord.

Take Rust for example, let's say we took the same code. This is `syn` but it's ~close enough to get the point:

```rust
fn main() {
    let expr: syn::Expr = syn::parse_str("a (1)").unwrap();
    println!("{:#?}", expr);
}
```

which gets us

```
Expr::Call {
    attrs: [],
    func: Expr::Path {
        attrs: [],
        qself: None,
        path: Path {
            leading_colon: None,
            segments: [
                PathSegment {
                    ident: Ident(
                        a,
                    ),
                    arguments: PathArguments::None,
                },
            ],
        },
    },
    paren_token: Paren,
    args: [
        Expr::Lit {
            attrs: [],
            lit: Lit::Int {
                token: 1,
            },
        },
    ],
}
```

So in that above example, the `args` is a `Lit::Int` of `1`, so the argument is an integer literal. So when `rustfmt` is formatting it, it's just outputting what the parser sees -- it's not doing any special handling for that argument type. It's just that the Rust parser allows spaces between parentheses and message names.

In Ruby, _this is not the same_. It you do `Prism.parse("a (1)")`, you get a different node type:

```
  @ ProgramNode (location: (1,0)-(1,5))
  ├── locals: []
[...]

PR #787, opened 2026-01-16
 make paren stripping work good

fables-tales:
This reverts commit f54d5af66eb9111796eff781b378a233c2b95607.

### DIFF
--- a/fixtures/small/paren_expr_preserve_expected.rb
+++ b/fixtures/small/paren_expr_preserve_expected.rb
@@ -38,7 +38,6 @@ foo(
     else
       :other
     end
-
   )
 )
 
@@ -50,6 +49,5 @@ foo(
     rescue
       fallback
     end
-
   )
 )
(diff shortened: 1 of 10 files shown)


fixtures/small/paren_expr_preserve_expected.rb
```diff
@@ -0,0 +1,55 @@
...
+# begin/end blocks
+foo(
+  (
+    begin
+      risky_operation
+    rescue
+      fallback
+    end
+
+  )
```
reese, line 52: What's up with the extra spacing here? I feel like we may be missing some line handling here or something

fixtures/small/paren_expr_preserve_expected.rb
```diff
@@ -0,0 +1,55 @@
...
+    case x
+    when 1
+      :one
+    when 2
+      :two
+    else
+      :other
+    end
+
+  )
```
reese, line 41: (Same here)

reese (commented):
Mostly LGTM but there's some weird trailing whitespace for case/begin blocks that I think is unintentional

reese (approved):

reese:
Closes #739

PR #792, opened 2026-01-16
Bump to `v0.12.0`

reese:
It's probably 'bout time we cut a release now that we've fully moved over to Prism -- this bumps us to `0.12.0`!


fables-tales (approved):

fables-tales (changes requested):
one sec actually

fables-tales (approved):

PR #795, opened 2026-01-17
fix #413

### LINKED ISSUE #413: Comment outside the block when it before has multiple empty lines.
* Ruby version: 3.2.0
* Rubyfmt : 0.8.1

## Input file

```ruby
Rails.application.configure do
  
  
  # Raises error for missing translations.
end
```

## Rubyfmt's output

```rb
Rails.application.configure do
end

# Raises error for missing translations.
```

## Expected

```rb
Rails.application.configure do
  # Raises error for missing translations.
end
```

fables-tales:
I let claude do it's thing

### DIFF
--- a/librubyfmt/src/parser_state.rs
+++ b/librubyfmt/src/parser_state.rs
@@ -484,15 +484,10 @@ impl<'src> ParserState<'src> {
                 && self
                     .comments_hash
                     .is_empty_line(self.current_orig_line_number + 1)
-                && self
-                    .comments_to_insert
-                    .as_ref()
-                    .is_some_and(|c| c.line_count() > 0)
+                && let Some(comments) = self.comments_to_insert.as_mut()
+                && comments.line_count() > 0
             {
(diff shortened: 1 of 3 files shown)


librubyfmt/src/parser_state.rs
```diff
@@ -475,19 +475,24 @@ impl<'src> ParserState<'src> {
...
                 .comments_hash
                 .has_line(self.current_orig_line_number + 1)
                 && self
                     .comments_hash
                     .is_empty_line(self.current_orig_line_number + 1)
-                && self.comments_to_insert.is_some()
+                && self
+                    .comments_to_insert
+                    .as_ref()
+                    .is_some_and(|c| c.line_count() > 0)
```
reese: _nit_ but I bet we can avoid that `expect` below with something like

```rust
&& let Some(comments) = self.comments_to_insert.as_mut() && comments.line_count() > 0
```

reese (approved):

PR #797, opened 2026-01-17
Fix --check mode to exit non-zero on syntax errors

### LINKED ISSUE #451: `--check [FILE]` returns `0` exit code for invalid syntax
* Rubyfmt git sha: `e00d2ab89fd4b0b85a7897fac393c1ad987136de` (0.10.0)

## Input file + Rubyfmt's output

```bash
$ cat hi.rb
bad syntax in here

$ cat hi.rb | cargo run -- --check
Rubyfmt detected a syntax error in the ruby code being executed

$ echo $?
1

$ cargo run -- --check hi.rb 
Rubyfmt detected a syntax error in the ruby code being executed

$ echo $?
0
```

## This breaks because

These two are inconsistent (despite having the same input), and realistically it should probably return `5` in both cases (since we return `5` for invalid syntax in normal runs).

fables-tales:
Previously, --check would exit with code 0 even when syntax errors were encountered in input files. This changes the check mode to track errors and exit with code 1 (SyntaxError) when any syntax errors occur.

Also adds tests for --check with syntax errors for both file and stdin input.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -349,8 +349,9 @@ fn main() {
             let text_diffs: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
             let errors_count: Arc<Mutex<usize>> = Arc::new(Mutex::new(0));
 
-            iterate_input_files(&opts, &|(file_path, before)| {
-                match rubyfmt_string(&opts, before) {
+            iterate_input_files(
+                &opts,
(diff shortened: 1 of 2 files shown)


reese (approved):

fables-tales:
fixes: https://github.com/fables-tales/rubyfmt/issues/451

PR #799, opened 2026-01-18
Speed up comment presence checks

reese:
This addresses two sources of slowness in our comment handling. First, `has_comment_in_offsets` is being stored in a `HashSet`, but in `has_comment_in_offsets` -- the only place we use this -- we just iterate through all of them anyways, making the `HashSet` redundant. We can use a `Vec` instead and avoid the hashing overhead.

And now that have the offsets in a hash, and because they're naturally sorted since we're inserting them in-order, we can then binary search for the offset instead of iterating through everything, which is O(log n) instead of O(n).


froydnj (approved):
Nice!

PR #800, opened 2026-01-18
speedup `has_comments_in_lines` with binary search

froydnj:
We don't hit this code very often, but we might as well try to be fast when we do.

This way performs a little bit more work than necessary -- there is probably a smarter way to do this based on binary searching for the start and then the end of the range, and then deciding based on those results, but this way is more elegant.

### DIFF
--- a/librubyfmt/src/file_comments.rs
+++ b/librubyfmt/src/file_comments.rs
@@ -1,4 +1,5 @@
 use memchr::memchr_iter;
+use std::cmp::Ordering;
 
 use crate::comment_block::CommentBlock;
 use crate::parser_state::line_difference_requires_newline;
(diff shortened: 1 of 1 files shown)


reese (approved):

froydnj:
Realized this morning that this is not quite right -- `contains` doesn't partition the slice in the way `partition_point` requires.  The partitioning is FTF instead of FT.  So we'll have to do the real binary search thing, which I'll write up later today.

fables-tales:
I have a concern that we're moving from a data structure that is inherently correct for this problem to one where we have to be careful about how we use it. Are we sure it's actually worth it here?

fables-tales (changes requested):
blocking this PR until it's resolved. Feel free to ask me to unblock if you're happy with it and have addressed the correctness question

fables-tales:
I'm going to close this PR for now, feel free to re-open it or file a new one if you want to update this feature

Issue #365, opened 2022-11-19
Support `--stdin-path` or something similar to properly ignore files formatted through stdin


reese:
This is more a feature request since I imagine other tools (e.g. VSCode extensions will need it, see the discussion in #354) will need it: currently, any tool formatting through stdin loses all ability to use `.rubyfmtignore`, but there are plenty of cases where reading from disk isn't convenient/possible, especially in VSCode extensions like the Ruby and Sorbet extensions.

Having something [like `--stdin-path` from `prettier`](https://prettier.io/docs/en/options.html#file-path) would simplify this problem by basically telling rubyfmt to format file contents as if they were at the given path, and ignore them as needed. This prevents extensions from having to manually handle `.rubyfmtignore` paths.

So for example, if my `.rubyfmtignore` has `foo.rb` in it, doing
```bash
echo 'puts "hi"' | rubyfmt --stdin-path foo.rb
# => puts "hi"
```
produces no changes due to the file being ignored, but
```bash
echo 'puts "hi"' | rubyfmt --stdin-path bar.rb
# => puts("hi")
```
produces changes as expected.

reese:
Hmm, so there's a bit of a snag on this, which is mainly that https://github.com/BurntSushi/ripgrep/issues/829 is still kicking around, which affects _exact_ use-case. We're also not the first Rust-based code formatter (see https://github.com/JohnnyMorganz/StyLua/issues/377) to get bitten by this, so I think this will likely turn out to be non-trivial (i.e. it may require hand-rolling out ignore logic, which sounds not-fun).

Gonna put this on the backburner for a bit, but it's worth noting that this also means that some pretty basic ignore patterns are gonna be busted for users, so this makes the `.rubyfmtignore` and `.gitignore` checking a bit less useful.

PR #805, opened 2026-01-19
Update `ignore` crate

reese:
https://github.com/BurntSushi/ripgrep/issues/829 causes the same issue in `rubyfmt` but has since been fixed. This adds test cases that fail without updating and pass afterwards, and then updates the crate.


froydnj (approved):

reese:
(These are the issues I was referring to in #365)

PR #809, opened 2026-01-19
Update issue template

reese:
This updates the issue template to reference https://rubyfmt.run (I've updated it to have a "Create Issue button"), which has a few benefits: it should make it faster to create issues in a consistent way, and it also regularly updates to `rubyfmt`'s `HEAD` commit, so it should prevent users from filing issues for bugs that are already fixed.

This template is basically the same as the [Sorbet](https://github.com/sorbet/sorbet/blob/master/.github/ISSUE_TEMPLATE/bug.md) issue template, which does the same thing for sorbet.run. It also moves it in to `.github/ISSUE_TEMPLATE/bug-report.md` with the newer markdown template format, since Github was saying that we were using a legacy issue template location.


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,47 @@
...
+that the file has changes.
+
+Please don't open "feature request" issues, or "I'd like to change the output to
+be more like this" issues. We've got a clear direction for those and we'll open
+issues up to those in a future release.
+ -->
+
+- `rubyfmt` version/commit: TODO <!--- If unsure, run `rubyfmt --version` --->
+
+[→ View on rubyfmt.run](TODO_replace_this_with_your_sorbet_run_link)
```
fables-tales: change sorbet to rubyfmt

fables-tales (changes requested):

fables-tales (approved):

PR #810, opened 2026-01-19
Reorganize the README

reese:
This rearranges the README to be more friendly to folks who don't already know what `rubyfmt` is. It leads with a one-liner about what it does, a link to rubyfmt.run to see it in action, and then tells them how to install and run it. It also updates the CLI flag docs in a bit more of a structured table format with separate sections for the different ignore mechanisms (headers and `.rubyfmtignore`/`.gitignore`).

The editor integrations are largely the same, though I did add a section for a `ruby-lsp` plugin since the old Ruby VSCode extension is now deprecated. I also put in a section on Rubocop since there were several open issues about it. (The ruby-lsp add-on and the rubocop-rubyfmt gem mentioned are mine, just plugging them since it doesn't seem like other community options exist/still work for these two cases.)

I also kept the pronunciation / maintenance policy here but just moved them to the bottom -- they're helpful to have, but not the first thing users need to see.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ A fast, opinionated Ruby formatter written in Rust.
 
 - Try it online → [rubyfmt.run](https://rubyfmt.run)
 
+Pronunciation: (en) "Ruby Format", (jp) ルビーフォーマット
+
 ## Quick Start
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,112 +1,167 @@
...
+Please check out our [contributing guide](./CONTRIBUTING.md).
+
+## Maintenance Status
+
+The original author (fables-tales) is no longer working on open source in their spare time.
+Other contributors work regularly on `rubyfmt`, and contributors with commit access are welcome to merge changes that pass CI.
+
+## FAQ
+
+**How do I pronounce rubyfmt?**
```
fables-tales: let's move this up top, otherwise I think this is good

fables-tales (changes requested):

fables-tales:
fixes https://github.com/fables-tales/rubyfmt/issues/424

fables-tales (approved):

PR #811, opened 2026-01-20
Fix `break` multilining

### LINKED ISSUE #806: multi-line `break` handling is not idempotent
Original input (with v11.253-0 and with HEAD), a modified version of `test_small_breaks`:

```ruby
loop do
  break no
  break
  break Some::Very::Long::Thing::That::Will::Extend::Over::Multiple::Lines::And::Never::Ends.await_result(workflow_id: workflow_id)
end
```

Running this through `rubyfmt` produces:

```ruby
loop do
  break no
  break
  break(
    Some::Very::Long::Thing::That::Will::Extend::Over::Multiple::Lines::And::Never::Ends.await_result(
      workflow_id: workflow_id
    )
  )
end
```

But running that through again produces:

```ruby
loop do
  break no
  break
  break(
    (Some::Very::Long::Thing::That::Will::Extend::Over::Multiple::Lines::And::Never::Ends.await_result(
      workflow_id: workflow_id
    ))
  )
end
```

reese:
Closes #806 

`break` in the grammar does not support being called with parens, so our multiline strategy is currently incorrect. I'm open to bikeshedding about how this should look (see the discussion in #806) but for the sake of un-breaking this, this PR formats multiline `break` calls the way we did historically, which is using `format_list_like_thing(... single_line: true)`.

### DIFF
--- a/fixtures/small/breaks_actual.rb
+++ b/fixtures/small/breaks_actual.rb
@@ -12,4 +12,9 @@ loop do
       attemptes: off_reservation_payment.attempts + 1
     )
   ]
+  break([
+    off_reservation_payment,
+    payment_intent_mode,
+    Return.new(outcome: Outcome::Succeeded, attemptes: off_reservation_payment.attempts + 1)
+  ])
 end
--- a/fixtures/small/breaks_actual.rb
+++ b/fixtures/small/breaks_actual.rb
@@ -1,4 +1,6 @@
 loop do
   break no
   break
+  break Some::Very::Long::Thing::That::Will::Extend::Over::Multiple::Lines::And::Never::Ends.await_result(workflow_id: workflow_id)
+  break Some::Very::Long::Thing::That::Will::Extend::Over::Multiple::Lines::And::Never::Ends.await_result(workflow_id: workflow_id), with_a_second_value
 end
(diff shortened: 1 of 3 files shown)


froydnj (changes requested):
Can you add testcases with multi-line arrays?  One of the (existing!) things I see in Stripe's codebase is something like:

```ruby
break(
  [
    off_reservation_payment,
    payment_intent_mode,
    Return.new(
      outcome: Outcome::Succeeded,
      attemptes: off_reservation_payment.attempts + 1
    )
  ]
)
```

I think it'd be good to have tests for calls with parens and slightly more complex arguments like this.

reese:
@froydnj What version of rubyfmt is that example from? I'm mostly curious how that even would be an output because if I run it through the last release before this (`v0.11.0`), that's not the output I see:

```bash
% cat example.rb
loop do
break(
  [
    off_reservation_payment,
    payment_intent_mode,
    Return.new(
      outcome: Outcome::Succeeded,
      attemptes: off_reservation_payment.attempts + 1
    )
  ]
)
end

% ~/Downloads/tmp/releases/v0.11.0-Darwin/rubyfmt example.rb

loop do
  break ([
    off_reservation_payment,
    payment_intent_mode,
    Return.new(
      outcome: Outcome::Succeeded,
      attemptes: off_reservation_payment.attempts + 1
    )
  ])
end
```

Or when you say "something like" that, are you saying the input is actually

```ruby
break [
  off_reservation_payment,
  payment_intent_mode,
  Return.new(
    outcome: Outcome::Succeeded,
    attemptes: off_reservation_payment.attempts + 1
  )
]
```

(Just trying to figure out what output you're expecting for that/how these examples should look.)

froydnj:
@reese It is from 0.11.250-0.  I guess I should have specified that the `break` starts pretty far-indented already; the real code is:

```ruby
              break [
                ...
```

(Ignore the wrapping parens; I forgot I was looking at a diff of prism output, not ripper output. 🤦‍♂️ )

froydnj (approved):
Thank you!

PR #816, opened 2026-01-21
don't unwrap parenthesized assignments in calls

froydnj:
Rubocop's `Lint/AssignmentInCondition` checks for code like:

```ruby
if var = thing
  # stuff
end
```

and complains at you.  A common way to work around this is to wrap the assignment in parens, like so:

```ruby
if (var = thing)
  # stuff
end
```

which indicates that, yes, I really mean to do this assignment here and I know what I'm doing.

(This is also a technique used in C/C++/etc.)

Stripe has a similar lint, but it checks for assignments in method call arguments:

```ruby
foo(var = thing)
```

would be disallowed.  Similarly to the above cop, wrapping the assignment in parens is the workaround:

```ruby
foo((var = thing))
```

This was allowed by the ripper implementation (accidentally?) -- i.e the parens were not removed -- and I think it is reasonable to do the same in the prism implementation.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -4943,6 +4943,9 @@ fn is_keyword_expression(node: &prism::Node) -> bool {
         Node::BeginNode { .. } => true,
         // For completeness: other control flow that shouldn't be unwrapped
         Node::ForNode { .. } => true,
+        // Assignments to local variables can be left wrapped for consistency with
+        // wrapping assignments in `if` conditions and the like.
+        Node::LocalVariableWriteNode { .. } => true,
         _ => false,
     }
 }
--- a/fixtures/small/paren_expr_unwrap_actual.rb
+++ b/fixtures/small/paren_expr_unwrap_actual.rb
@@ -37,3 +37,9 @@ foo (1..10)
 
 # Ternary without keywords
 foo (condition ? a : b)
+
+# Assignments we leave alone, for consistency with code like:
+#
+# if (var = thing)
+foo ((var = thing))
+foo(1, (var2 = thing))
--- a/fixtures/small/paren_expr_unwrap_expected.rb
+++ b/fixtures/small/paren_expr_unwrap_expected.rb
@@ -37,3 +37,9 @@ foo(1..10)
 
 # Ternary without keywords
 foo(condition ? a : b)
+
+# Assignments we leave alone, for consistency with code like:
+#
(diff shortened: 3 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -4943,6 +4943,9 @@ fn is_keyword_expression(node: &prism::Node) -> bool {
         Node::BeginNode { .. } => true,
         // For completeness: other control flow that shouldn't be unwrapped
         Node::ForNode { .. } => true,
+        // Assignments to local variables can be left wrapped for consistency with
+        // wrapping assignments in `if` conditions and the like.
+        Node::LocalVariableWriteNode { .. } => true,
```
reese, line 4948: I actually don't know, so checking -- are there any other write node types you need here? Like do we care about `foo((bar.baz = thing))` (which is a different node type)? I don't think that's the most common use case since I think usually people do `if (foo = bar)` for nil checking, but curious if we should have it for completeness.

reese (approved):

librubyfmt/src/format_prism.rs
froydnj: Eyeballing the rubocop failures and the lint in Stripe's codebase, the lint specifically checks only for local variables.

I am of two minds here: yes, we probably could check for instance variables, class variables, multiple assignments, and so forth, and that would be consistent.  However, my gut feeling (as well as a little bit of laziness) says that local variables are the overwhelmingly common thing.  A quick:

```
rg '(, |\()@[a-z0-9_+] = ' vendor/bundle
```

searching through hundreds of Stripe's bundled gems (a decent but perhaps not complete cross-section of Ruby gems) suggests that non-local variable assignments in calls -- or even in `if` conditions and the like, since the paren there should match that case -- aren't super common.

froydnj: OK, I can't write a regex correctly; looking for `rg '(, |\()@[a-z0-9_]+ = ' vendor/bundle` turns up five instances of people writing `if (@var = thing)`.  But nothing in a call argument, so I feel a little better about limiting here.

reese: Makes sense -- though I do wonder if in that case you also want an `InstanceVariableWriteNode` in there, I guess depends on if that lint fails for that as well. But I'll leave that up to you, 5 instances in vendor/bundle seems like it's ~very rare

PR #823, opened 2026-01-27
[perf] Use slices instead of `HashSet` for reserved method name checks

reese:
At first glance, `HashSet` made sense for fast checking of set membership, but for sets that are only ever 2-4 items, array iteration is significantly faster than the hashing overhead required for `HashSet::contains`. On my machine, `contains` on a slice of 2-4 items is about 10-12x faster than HashSets of the equivalent size.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, collections::HashSet, sync::LazyLock};
+use std::borrow::Cow;
 
 use ruby_prism as prism;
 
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Did your profile show locking overhead or just general overhead smeared over the whole lookup process?

reese:
In the profiles for a full formatting run, I didn't see any locking-specific samples. The benchmark mentioned in the description is more looking at HashSet vs slices in isolation (just benchmarking the two methods without a `LazyLock`), so it's possible there was also more overhead from the lock too, but this is enough of a micro-optimization that it just shows up as general overhead in profiles for a full run.

PR #826, opened 2026-01-27
Fix empty comment handling

reese:
`render_as` is pretty slow, largely due to the pattern matching required for a whole lot of band-aid fixes to what really should be fixed elsewhere. I'd like to remove most of these by just fixing the busted logic instead of applying band-aids, which hopefully will speed up `render_as` as a nice side effect.

---

In the line winding machinery, we try and maintain user-inserted newlines between comments and nodes. I don't recall why we don't use newlines, but for whatever reason we use empty `Comment` nodes (e.g. `Comment { comment: "" }`). The `into_line_comments` machinery was [_supposed_ to](https://github.com/fables-tales/rubyfmt/blob/cadd3d90081d37efbb7ffbc60cdda371624e1cf2/librubyfmt/src/comment_block.rs#L39-L40) ignore these but didn't, causing us to add additional hard newlines in between these empty comments. This updates the logic in `into_line_comments` to ignore all but the trailing empty comment lines (which are the ones actually used for spacing) and then removes the old cleanup logic.

### DIFF
--- a/librubyfmt/src/comment_block.rs
+++ b/librubyfmt/src/comment_block.rs
@@ -25,11 +25,30 @@ impl CommentBlock {
     }
 
     pub fn into_line_tokens<'src>(self) -> impl Iterator<Item = ConcreteLineToken<'src>> {
-        self.comments.into_iter().flat_map(|c| {
-            [
-                ConcreteLineToken::Comment { contents: c },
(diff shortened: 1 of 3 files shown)


froydnj (approved):
Fully approve of fixing things earlier in the pipeline if possible.

PR #831, opened 2026-02-27
Clean up redundant Cow usage in variables scopes

reese:
This `Cow` usage is a vestigial change from supporting both `String` and `&str` when Ripper was still around, so we can 🔪 it


froydnj (approved):
Nice.

Issue #832, opened 2026-02-27
Files not encoded with valid UTF-8 cannot be formatted


reese:
**Commit:** [`dfd2442`](https://github.com/fables-tales/rubyfmt/commit/dfd2442c4d88fc88cfc0fbd0d2908c7308fed969)

#### Input

```ruby
# encoding: EUC-JP                                                                                                                                                                                                       

# 顧客管理システム

def 挨拶(名前)
  "こんにちは、#{名前}さん"
end
```

(Though note that it needs to be in the proper encoding, so to generate an example file, try this:)

```ruby
content = "# encoding: EUC-JP\n\ndef 挨拶(名前)\n  \"こんにちは、\#{名前}さん\"\nend\n"
File.write("hi.rb", content.encode("EUC-JP"), encoding: "binary")
```

#### Output

(Error!)

#### Error

```
Encountered a syntax error
```

#### Expected behavior

We should be able to format non-UTF-8 files. In [a few places](https://github.com/search?q=repo%3Afables-tales%2Frubyfmt+from_utf8+-path%3Atests%2F&type=code), we use the `str`/`String::from_utf8` utilities to read bytes into strings, but in general we should prefer operating directly on byte arrays instead. (This will also help performance, as currently about 3% of our runtime is spent on UTF-8 validation.)

PR #835, opened 2026-02-28
Make `format_buffer` return a `Vec<u8>`

reese:
`format_buffer` currently returns a `String` and assumes that we always return valid UTF-8. With the changes I'm making for #832, this will no longer be the case, since Ruby source is not always UTF-8 encoded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ ctrlc = { version = "3.2", features = ["termination"] }
 ignore = "0.4.18"
 regex = "1.11.1"
 rubyfmt = { path = "./librubyfmt" }
-similar = "2.1.0"
+similar = { version = "2.7.0", features = ["bytes"] }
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -17,7 +17,7 @@ ctrlc = { version = "3.2", features = ["termination"] }
 ignore = "0.4.18"
 regex = "1.11.1"
 rubyfmt = { path = "./librubyfmt" }
-similar = "2.1.0"
+similar = { version = "2.7.0", features = ["bytes"] }
```
reese, line 20: The update to `2.7.0` is just to keep us current with the latest version -- it doesn't have any functional differences as far as I can tell, just a bit of 🧹 

The `bytes` feature is so that we can pass u8 Vecs to `TextDiff::from_lines`, which by default only handles strings.

froydnj (approved):

PR #840, opened 2026-03-04
Remove fancy-regex dependency

reese:
Currently, any time we convert a string to use double quotes, we compile and run a regex for every string. Benchmarking the large test fixtures, `Regex::new` alone takes up more than 2% of the overall runtime, with `single_to_double_quoted` taking up 3.5%. We could cache these or precompile the most common ones, but honestly using a regex here is overkill -- we're basically just checking for slashes followed by one of the delimiters, which we could do by iterating over the characters instead.

Replacing the `fancy-regex` usage with a manual implementation shows `single_to_double_quoted` going from 3.5% -> 0.5% in my benchmarking, plus it's one fewer dependency.

(This is also part of #832, since `fancy-regex` takes UTF-8 strings.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,21 +119,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bit-set"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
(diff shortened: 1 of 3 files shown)


froydnj (approved):
I assume we have good test coverage for this?

reese:
Yeah, this should be covered pretty thoroughly by the [string literals stress test](https://github.com/fables-tales/rubyfmt/blob/trunk/ci/string_literals_stress_test.rb)

PR #841, opened 2026-03-17
Update CLI machinery to use `&[u8]` instead of strings

reese:
Part of #832 

This is mostly updating `read_to_string` to use `read` instead, and then updating anything downstream of that as needed.


froydnj (approved):
Yes!

PR #842, opened 2026-03-17
Finish support for non-UTF-8 source files

### LINKED ISSUE #832: Files not encoded with valid UTF-8 cannot be formatted

reese:
Closes #832 

This updates all of the string conversion internals to operate on `Vec<u8>` instead of `String`. This is a bit of a big change, but it's all intertwined and is easiest to change all at once, and the change is generally mechanical. This also gets rid of the last usages of `loc_to_str` and `u8_to_str` and removes them both.

### DIFF
--- a/fixtures/small/euc_jp_encoding_actual.rb
+++ b/fixtures/small/euc_jp_encoding_actual.rb
@@ -1,0 +1,5 @@
+# encoding: EUC-JP
+
+def 挨拶(名前)
+  "こんにちは、#{名前}さん"
+end
--- a/fixtures/small/euc_jp_encoding_expected.rb
+++ b/fixtures/small/euc_jp_encoding_expected.rb
@@ -1,0 +1,5 @@
+# encoding: EUC-JP
+
+def 挨拶(名前)
+  "こんにちは、#{名前}さん"
+end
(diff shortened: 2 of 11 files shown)


froydnj (approved):
Neat!

PR #843, opened 2026-03-17
🧹  Remove some elidable lifetimes

reese:
A while back I added a ton of lifetimes to the main formatting code since it was needed to pass around references to Prism nodes, but some of them aren't truly necessary, since they don't circle back into formatting code and are just helpers.


froydnj (approved):
I love PRs where I can just trust the compiler to do the review for me.

PR #844, opened 2026-03-17
Add unit tests for string escape logic

reese:
`rubyfmt` is almost entirely tested with fixtures. This is mostly fine, but there's a few isolated bits of logic that could use some unit testing. String escaping is the most obvious one -- it's bytes in to bytes out and has quite specific rules that are easy to test in isolation. (Heredocs and some of the comment bits also could use them, but strings are probably the highest value for now.)

This does add one `else` branch to the actual production code purely as a defense mechanism/future-proof that I saw in unit tests, but I'm fairly confident it's never reachable in real Ruby code (see the `word_array_trailing_backslash_preserved` test). It covers a trailing slash in a `%w` array, which in real usages (`%w[\]`) would escape the closing delimiter and be invalid, so that `else` is unreachable, but as far as `escape_word_array_content` is concerned it does the right thing if it ever were to get in that state.

### DIFF
--- a/librubyfmt/src/string_escape.rs
+++ b/librubyfmt/src/string_escape.rs
@@ -115,6 +115,8 @@ pub fn escape_word_array_content(
                     output.push(next);
                 }
                 bytes.next();
+            } else {
+                output.push(b'\\');
             }
         } else if c == TARGET_OPEN || c == TARGET_CLOSE {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Yes please, thank you for doing this.

PR #852, opened 2026-03-26
Remove `make_source_release`

reese:
We've always had this special step in the release process to add all of our source code as a release artifact, but GitHub does all of this for you by default when you make a release, so it's just duplicate work. This probably made more sense when we had a Ruby submodule and might have been trying specifically to include the Ruby source, but I don't think we need it anymore.


froydnj (approved):

fables-tales:
yes, the rationale for this was getting the ruby build in, which github did
not do, we can remove this now

PR #858, opened 2026-05-06
Fix interpolated symbol shorthand hash keys producing invalid Ruby

### LINKED ISSUE #857: Hash with interpolated shorthand symbol key (`{ "#{x}": 1 }`) is rewritten to invalid Ruby `{:"#{x}": 1}`
- `rubyfmt` version/commit: 0.13.0 (built from `trunk` at 8ad8723)

## Input

[→ View on rubyfmt.run](https://rubyfmt.run/#eyAiI3t4fSI6IDEgfQ==)

```ruby
{ "#{x}": 1 }
```

#### Output

```ruby
{:"#{x}": 1}

```

## Expected behavior

The shorthand symbol-key form with interpolation should probably be preserved as `{"#{x}": 1}` (still a valid hash with a symbol key, equivalent to the input). What rubyfmt produces today is invalid Ruby — the leading `:` is added on top of the trailing shorthand `:`, so a second pass through rubyfmt fails with `Rubyfmt detected a syntax error in the ruby code being executed`. `ruby -c` on the formatted output reports `unexpected ':'`.

It looks like rubyfmt is normalizing the key to a `:"..."` symbol-literal form but not also dropping the trailing `:` that comes from the shorthand syntax.

---

Reproduction:

```sh
echo '{ "#{x}": 1 }' > /tmp/repro.rb
rubyfmt -i /tmp/repro.rb   # rewrites to `{:"#{x}": 1}` — already invalid Ruby
rubyfmt -i /tmp/repro.rb   # second pass: "Rubyfmt detected a syntax error..."
```

Originally hit on a real file containing `{ "#{address_type}": address.previous_changes.except('updated_at') }.merge(...)`.

dduugg:
Closes #857.

## Bug

A hash with an interpolated shorthand symbol key was rewritten to invalid Ruby:

```ruby
{ "#{x}": 1 }   # input
{:"#{x}": 1}    # rubyfmt output — invalid; second pass fails as a syntax error
```

## Cause

`format_interpolated_symbol_node` hard-coded the opener as `:"` and the closer as `"`, ignoring prism's `opening_loc` / `closing_loc`. That works for the standalone literal `:"#{x}"` but corrupts the shorthand-key form, where prism reports `opening_loc = "` and `closing_loc = ":`. The function emitted `:"#{x}"`, then `format_assoc_node` appended its own `:` separator.

## Fix

Mirror the pattern that `format_symbol_node` already uses for plain `SymbolNode`s: emit the actual opening / closing locations and strip a trailing `:` from the closer (since `format_assoc_node` re-emits that as the key/value separator). A `debug_assert!` documents that the locations are always present in practice; the previous `:"` / `"` defaults remain as a graceful release-mode fallback.

## Test

Fixtures cover:

- `{ "#{x}": 1 }` — bare hash, the original failure mode
- `foo("#{x}": 1)` — keyword-arg call site (distinct caller of `format_assoc_node`)
- `{ "#{x}": 1, "#{y}": 2 }` — multi-entry, distinguishes a per-first-key fix from a general one
- `{ a: 1, "#{x}": 2 }` — mixed plain-shorthand and interpolated-shorthand

The existing `:"#{name}="` literal coverage in `string_first_item_is_embed` still passes (regression check). `cargo fmt --check` and `cargo clippy -D warnings` are clean.

## Follow-on refactor

Per the suggestion to consider centralizing the mirrored strip logic, commit `4d00438` extracts `emit_symbol_key_closer` next to `format_assoc_node`. Both `format_symbol_node` and `format_interpolated_symbol_node` now route through it, so the `strip_suffix(b":")` pattern lives in one place with a doc comment naming `format_assoc_node` as the owner of the `:` separator. No behavior change — verified by the full fixtures suite (379 tests) and an extra round-trip on the mixed-form path (`{foo: 1, 'bar' => 2}` → `{:foo => 1, "bar" => 2}`), which exercises the strip in the rocket-rendering branch where `as_symbol=false`.


reese (approved):
LGTM. I think a refactor would also be fine if you wanted to move some of that mirrored logic into `format_assoc_node` (in this PR or in a follow-on), up to you.

dduugg:
@reese TY for the quick review. I added a second commit to de-duplicate the logic, PTaL.

PR #863, opened 2026-05-08
Bump `prerelease` levels instead of `prepatch` in `preview-release` GH action

reese:
This is something minor that has annoyed me for a long time: in the `preview-release` action, we currently bump using the `prepatch` level, which always increments the `patch` semver level. This makes it really annoying to properly release a true patch bump, since every commit is bumping the patch. For example, I would love to do a patch release right now for some small fixes, but that would require us to bump from `0.13.0` to `0.13.8`, since we've been bumping the patch version this whole time.

Instead, we should be using `prerelease`, which will bump from e.g. `0.13.0-1` to `0.13.0-2`, and when we do make a release, say `0.13.1`, it bumps to `0.13.2-0`, which importantly does not prevent us from releasing `0.13.2` itself.

### DIFF
--- a/.github/workflows/preview-release.yaml
+++ b/.github/workflows/preview-release.yaml
@@ -16,7 +16,7 @@ jobs:
         id: bump-semver
         with:
           current_version: ${{ steps.get-latest-tag.outputs.tag }}
-          level: prepatch
+          level: prerelease
       - uses: actions-ecosystem/action-push-tag@v1
         with:
           tag: ${{ steps.bump-semver.outputs.new_version }}


froydnj (approved):
I kind of like the current scheme, but I understand that it's a bit painful for actual patch releases.  Releases are cheap, though, we could just do 0.14.0?

reese:
Yeah in this particular case I don't care about the next one being `0.14.0`, but long-term I think it's probably not great that we basically cannot do patch releases without this, and conceptually I find it easier to grok that `0.13.1-*` prereleases include all changes that could eventually be in the proper `0.13.1` release.

For the most part, historically we've just completely avoided any proper semantic versioning, but since things are slightly more stable now, it might be nice to actually have a distinction between major/minor/patch in future versions.

froydnj:
0.x versions forever! /s

fables-tales:
@reese the next release that feels like it has a good chunk of stuff let's just call it 1.0

reese:
1.0 would probably be a great one to align with Prism 2.0 (https://github.com/ruby/prism/issues/3990), which will have some nice stuff for us (arena allocation!)

fables-tales:
sgtm

PR #872, opened 2026-05-10
Fix stabby lambdas dropping semicolons in multi-statement bodies

thomasklemm:
## Summary

Stabby lambdas (`->`) with multiple statements separated by semicolons were formatted inline, silently dropping the semicolons and producing invalid Ruby code.

**Root cause:** `format_lambda_node` used `emit_soft_newline()` between statements inside an `inline_breakable_of` block. Soft newlines get collapsed when the breakable stays inline, so the semicolons simply disappeared. Regular brace blocks (`format_block_node`) already handled this correctly by forcing hard newlines when there are multiple statements.

## Examples

Before (broken):
```ruby
-> { a; b }              # becomes: -> { a b }
->(x) { puts x; x + 1 } # becomes: -> (x) { puts(x) x + 1 }
```

After (correct):
```ruby
-> { a; b }
# becomes:
-> {
  a
  b
}

->(x) { puts x; x + 1 }
# becomes:
-> (x) {
  puts(x)
  x + 1
}
```

Single-statement stabby lambdas remain inline as before:
```ruby
-> { a } # stays: -> { a }
```

## Test plan

- [x] Added `fixtures/small/stabby_lambda_multi_stmt_actual.rb` / `_expected.rb` covering multi-statement and single-statement cases
- [x] Verified test fails without the fix (produces `-> { a b }`)
- [x] Verified test passes with the fix
- [x] Full test suite passes (383 fixture + 46 CLI + 11 error handling + 3 stress tests)

### DIFF
--- a/fixtures/small/stabby_lambda_multi_stmt_actual.rb
+++ b/fixtures/small/stabby_lambda_multi_stmt_actual.rb
@@ -1,0 +1,9 @@
+-> { a; b }
+
+->(x) { puts x; x + 1 }
+
+->(p) { p.id; {kpis: {}, daily_summary: []} }
+
+x = -> { first; second; third }
+
+-> { only_one }
--- a/fixtures/small/stabby_lambda_multi_stmt_expected.rb
+++ b/fixtures/small/stabby_lambda_multi_stmt_expected.rb
@@ -1,0 +1,22 @@
+-> {
+  a
+  b
+}
+
+-> (x) {
+  puts(x)
+  x + 1
+}
+
+-> (p) {
+  p.id
+  {kpis: {}, daily_summary: []}
+}
+
+x = -> {
+  first
+  second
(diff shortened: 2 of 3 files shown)


thomasklemm:
Note: This PR has two commits that each pass the full test suite independently:

1. **The minimal fix** — adds multi-statement handling to `format_lambda_node`, mirroring the existing logic in `format_block_node`
2. **Optional refactoring** — extracts the shared logic into a `format_brace_block_body` helper, deduplicating the body-formatting code between the two functions and replacing `is_some_and` + `.unwrap()` with safer `if let` patterns

Feel free to squash both, drop the second commit if you prefer the fix without the refactoring, or let me know if you'd like any changes.

reese (approved):
LGTM, thanks for working on this!

PR #876, opened 2026-05-10
[fix] Format singleton class receiver expression instead of hardcoding self

### LINKED ISSUE #874: class << expr silently rewrites receiver to self
## Summary

rubyfmt silently rewrites the receiver of `class << expr` to `self`, regardless of what `expr` is. This changes the **semantics** of the code — it opens the singleton class of a completely different object.

## Reproduction

```ruby
# Input
class << receiver; self; end
class << obj; self; end
class << some_object
  def foo; :bar; end
end

# rubyfmt output (ALL become class << self)
class << self
  self
end

class << self
  self
end

class << self
  def foo
    :bar
  end
end
```

## Impact

This is a **silent semantic change**, not a formatting issue. The formatted code parses and runs without errors, but operates on the wrong object.

In our codebase, a `class << receiver` in a test stub helper was rewritten to `class << self`, causing **319 test failures** (`TypeError: can't bind singleton method to a different class`). Without a test suite, this would have shipped undetected.

## Environment

- rubyfmt 0.13.0 (Homebrew, macOS arm64)
- Also reproduced in a local build from current main

thomasklemm:
## Summary

- `format_singleton_class_node` emitted a hardcoded `self` identifier regardless of the actual receiver expression, silently rewriting `class << obj` to `class << self`
- Now formats the AST expression node directly using `format_node`, preserving whatever receiver the user wrote
- Added fixture test covering local variables, `self`, constants, and instance variables as receivers

Fixes #874

## Examples

Before (broken):
```ruby
class << receiver; self; end      # becomes: class << self\n  self\nend
class << obj; self; end           # becomes: class << self\n  self\nend
class << some_object              # becomes: class << self
  def foo; :bar; end              #            def foo; :bar; end
end                               #          end
```

After (correct):
```ruby
class << receiver
  self
end

class << obj
  self
end

class << some_object
  def foo
    :bar
  end
end
```

## Test plan

- [x] Added `fixtures/small/singleton_class_expression_actual.rb` / `_expected.rb` covering 6 receiver variants
- [x] Verified test fails without the fix (all receivers become `self`)
- [x] Verified test passes with the fix
- [x] Full test suite passes (383 fixture + 46 CLI + 11 error handling + 3 stress tests)

### DIFF
--- a/fixtures/small/singleton_class_expression_actual.rb
+++ b/fixtures/small/singleton_class_expression_actual.rb
@@ -1,0 +1,31 @@
+class << self
+  def foo; end
+end
+
+class << receiver
+  def foo; :bar; end
+end
+
+class << @ivar
+  def foo; end
+end
+
+class << @@cvar
+  def foo; end
+end
+
+class << $global
+  def foo; end
+end
+
+class << Foo
+  def foo; end
+end
+
+class << SomeModule::CONST
+  def foo; end
+end
+
+class << foo.bar
+  def foo; end
+end
--- a/fixtures/small/singleton_class_expression_expected.rb
+++ b/fixtures/small/singleton_class_expression_expected.rb
@@ -1,0 +1,40 @@
+class << self
+  def foo
+  end
+end
+
+class << receiver
+  def foo
+    :bar
+  end
+end
+
+class << @ivar
+  def foo
+  end
+end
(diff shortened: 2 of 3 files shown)


reese (approved):
🚀

PR #882, opened 2026-05-10
[fix] Preserve inline directive comments instead of moving them above

### LINKED ISSUE #881: Moving inline rubocop:disable to line above silently widens its scope
## Summary

When rubyfmt moves an inline `# rubocop:disable` comment to the line above, it silently changes the disable scope from **one line** to **the rest of the file**. This is not an aesthetic issue — it weakens linting and security checks without any warning.

Related: #260 (closed as known limitation), #866 (general discussion about comment repositioning). This issue focuses specifically on the **semantic correctness** impact that goes beyond comment aesthetics.

## Reproduction

[→ View on rubyfmt.run](https://rubyfmt.run/#Y2xhc3MgVXNlckNvbnRyb2xsZXIgPCBBcHBsaWNhdGlvbkNvbnRyb2xsZXIKICBkZWYgcHJvbW90ZQogICAgdXNlci51cGRhdGVfY29sdW1uKDpyb2xlLCAiYWRtaW4iKSAjIHJ1Ym9jb3A6ZGlzYWJsZSBSYWlscy9Ta2lwc01vZGVsVmFsaWRhdGlvbnMKICAgIHVzZXIudXBkYXRlX2NvbHVtbig6dmVyaWZpZWQsIHRydWUpCiAgZW5kCmVuZA==)

```ruby
# Input
class UserController < ApplicationController
  def promote
    user.update_column(:role, "admin") # rubocop:disable Rails/SkipsModelValidations
    user.update_column(:verified, true)
  end
end

# rubyfmt output
class UserController < ApplicationController
  def promote
    # rubocop:disable Rails/SkipsModelValidations
    user.update_column(:role, "admin")
    user.update_column(:verified, true)
  end
end
```

**Before:** Only the first `update_column` is exempt from the cop. The second line is still checked.

**After:** Both `update_column` calls (and everything else until end of file) are exempt, because a `# rubocop:disable` on its own line without a matching `# rubocop:enable` disables the cop for the remainder of the file.

## Why this is different from #866

In #866, the discussion is about comment aesthetics — whether a comment's *intended association* with a line is preserved. That's a style judgment call.

`rubocop:disable` is not a comment for humans. It is a **machine-readable directive** with formally defined scoping rules:

- **Inline** (end of a code line): disables the cop for that line only
- **Own line** (without a matching `enable`): disables the cop from that point to end of file

Moving the comment from inline to its own line **changes its semantics**. The author deliberately chose inline placement to limit the scope. rubyfmt silently widens it.

## Security-relevant examples

```ruby
# These inline disables are intentionally narrow-scoped:
render html: content.html_safe # rubocop:disable Rails/OutputSafety
eval(user_input) # rubocop:disable Security/Eval
```

After formatting, these cops are disabled for the rest of the file.

## Precedent in other formatters

Other formatters treat linter directives as a special case:

- **Black** (Python) preserves `# type: ignore`, `# noqa`, `# pragma: no cover` inline — they are never moved
- **Prettier** (JavaScript) preserves `// eslint-disable-line` inline
- **gofmt** (Go) preserves `//nolint` inline

These tools recognize that linter directives have **formal semantics** that differ from regular comments.

## Suggested approach

A targeted fix that detects directive-pattern comments and preserves them inline:

```
rubocop:disable, rubocop:enable, rubocop:todo
steep:ignore
type: ignore (Sorbet)
```

This would be a narrow, well-scoped change — only comments matching known directive patterns need inline preservation, not all trailing comments (which is the harder general problem discussed in #866).

## Environment

- rubyfmt 0.13.0 (Homebrew, macOS arm64)
- RuboCop 1.76

thomasklemm:
Closes #881

## Problem

Trailing linter directive comments were moved to their own line above the code, silently changing their semantics.

In RuboCop, placement determines scope ([docs](https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code)):
- **Trailing** `code # rubocop:disable Cop` → disables for **that line only** (via `DirectiveComment#single_line?`)
- **Standalone** `# rubocop:disable Cop` → disables **until matching `enable` or EOF**

rubyfmt was converting every trailing directive into a standalone one — silently widening the disable scope to the rest of the file. This is especially problematic for security-sensitive cops like `Rails/OutputSafety` and `Security/Eval`.

[→ See the bug on rubyfmt.run](https://rubyfmt.run/#Y2xhc3MgVXNlckNvbnRyb2xsZXIgPCBBcHBsaWNhdGlvbkNvbnRyb2xsZXIKICBkZWYgcHJvbW90ZQogICAgdXNlci51cGRhdGVfY29sdW1uKDpyb2xlLCAiYWRtaW4iKSAjIHJ1Ym9jb3A6ZGlzYWJsZSBSYWlscy9Ta2lwc01vZGVsVmFsaWRhdGlvbnMKICAgIHVzZXIudXBkYXRlX2NvbHVtbig6dmVyaWZpZWQsIHRydWUpCiAgZW5kCgogIGRlZiBzaG93CiAgICByZW5kZXIgaHRtbDogY29udGVudC5odG1sX3NhZmUgIyBydWJvY29wOmRpc2FibGUgUmFpbHMvT3V0cHV0U2FmZXR5CiAgZW5kCgogIGRlZiBoYXNfcm9sZT8obmFtZSkgIyBzdGFuZGFyZDpkaXNhYmxlIE5hbWluZy9QcmVkaWNhdGVNZXRob2QKICAgIHJvbGVzLmluY2x1ZGU/KG5hbWUpCiAgZW5kCgogIGRlZiBjb21wdXRlCiAgICByZXN1bHQgPSB1bnR5cGVkX21ldGhvZCAjIHN0ZWVwOmlnbm9yZQogICAgeCA9IDQyICMganVzdCBhIHJlZ3VsYXIgY29tbWVudAogIGVuZAplbmQ=)

## Solution

Detect known directive patterns and preserve them inline. Non-directive trailing comments continue to move above (rubyfmt's intended behavior).

**Before:**
```ruby
# rubocop:disable Rails/SkipsModelValidations  ← now file-scoped!
user.update_column(:role, "admin")
```

**After:**
```ruby
user.update_column(:role, "admin") # rubocop:disable Rails/SkipsModelValidations  ← line-scoped ✓
```

### Recognized patterns

`rubocop:disable/enable/todo/push/pop`, `standard:disable/enable`, `steep:ignore`, `:reek:`

## Implementation

The comment pipeline has three stages; this PR touches each one:

1. **Classification** (`file_comments.rs`): `is_inline_directive()` detects directive patterns. During comment extraction, directives on code lines are separated from regular comments. Also prevents trailing comments on code lines from entering the start-of-file comment sled.

2. **Token** (`line_tokens.rs`): New `InlineComment` variant renders as ` # contents` (space-prefixed, no trailing newline) — unlike `Comment` which always gets its own line.

3. **Emission** (`parser_state.rs`): Directives are buffered in `pending_inline_comments` and flushed before each `HardNewLine` in `emit_newline()`. Inside delimiter expressions (argument lists, arrays), directives fall back to the existing behavior (moved above) to avoid drifting to the closing bracket.

## Test plan

- [x] Added `fixtures/small/inline_directive_comments` covering: all 9 directive patterns, standalone vs trailing, mixed with regular comments, heredocs, multiline calls, conditionals, rescue, case/when, blocks, indented classes, constant assignments
- [x] Verified existing fixtures pass unchanged (including `paren_expr_calls`)
- [x] Verified idempotency
- [x] Full suite: 386 fixture + 46 CLI + 11 error + 3 stress tests
- [x] `cargo fmt` + `cargo clippy` clean

### DIFF
--- a/fixtures/small/paren_expr_calls_expected.rb
+++ b/fixtures/small/paren_expr_calls_expected.rb
@@ -3,10 +3,9 @@ other_cool_method((a + b).round(4))
 
 # rubocop:disable PrisonGuard/PrivateModule
 (foo(
-  # rubocop:enable PrisonGuard/PrivateModule
   foo
 ))
-  .flatten
+  .flatten # rubocop:enable PrisonGuard/PrivateModule
 
 # rubocop:disable Style/Stuff
 (MyModel::InSomeNamespace
--- a/fixtures/small/paren_expr_calls_expected.rb
+++ b/fixtures/small/paren_expr_calls_expected.rb
@@ -3,9 +3,10 @@ other_cool_method((a + b).round(4))
 
 # rubocop:disable PrisonGuard/PrivateModule
 (foo(
+  # rubocop:enable PrisonGuard/PrivateModule
   foo
 ))
-  .flatten # rubocop:enable PrisonGuard/PrivateModule
+  .flatten
 
 # rubocop:disable Style/Stuff
 (MyModel::InSomeNamespace
--- a/fixtures/small/inline_directive_comments_actual.rb
+++ b/fixtures/small/inline_directive_comments_actual.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+# Basic rubocop:disable - must stay inline
+user.update_column(:role, "admin") # rubocop:disable Rails/SkipsModelValidations
+user.update_column(:verified, true)
+
+# rubocop:enable - must stay inline
+x = eval(input) # rubocop:enable Security/Eval
+
+# rubocop:todo - must stay inline
+x = 1 # rubocop:todo Style/FrozenStringLiteralComment
+
+# Multiple cops on one directive
+foo(bar) # rubocop:disable Style/Foo, Style/Bar
+
+# Directive with explanation text
+SolarPark.where(id: ids) # rubocop:disable Tenant/UnscopedModelAccess -- IDs pre-filtered by org
+
+# standard:disable - must stay inline
+def has_role?(name) # standard:disable Naming/PredicateMethod
+  roles.include?(name)
+end
+
+# standard:enable - must stay inline
+def valid? # standard:enable Naming/PredicateMethod
+  true
+end
+
+# steep:ignore - must stay inline
+result = untyped_method # steep:ignore
+
+# :reek: - must stay inline
+def complex_method(a, b, c, d, e) # :reek:LongParameterList
+  a + b + c + d + e
+end
+
+# Regular trailing comment - must still move above
+x = 42 # just a note
+y = 99 # another comment
+
+# Standalone directive comment on own line - stays on own line
+# rubocop:disable all
+def dangerous_method
+  eval("puts 1")
+end
+# rubocop:enable all
+
+# Directive inside a block
+[1, 2, 3].each do |i|
+  puts i # rubocop:disable Rails/Output
+end
+
+# Directive on method call with parens added by rubyfmt
+render html: content.html_safe # rubocop:disable Rails/OutputSafety
+
+# Directive inside conditional
+if condition
+  do_something # rubocop:disable CustomCop/Whatever
+else
+  do_other # rubocop:disable CustomCop/Other
+end
+
+# Indented directive in a class
+class MyClass
+  def foo
+    bar # rubocop:disable Style/Foo
+  end
+end
+
+# Multiple lines with directives and regular comments
+a = 1 # regular comment
+b = 2 # rubocop:disable Style/NumericLiterals
+c = 3 # another regular comment
+d = 4 # rubocop:enable Style/NumericLiterals
+
+# Heredoc with directive
+content = <<~HEREDOC # rubocop:disable Style/HeredocStyle
+  hello world
+HEREDOC
+
+# Multiline method call with directive on closing line
+result = very_long_method(
+  arg1,
+  arg2,
+  arg3
+) # rubocop:disable Metrics/ParameterLists
+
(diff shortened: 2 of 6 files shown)


fixtures/small/paren_expr_calls_expected.rb
```diff
@@ -3,10 +3,9 @@
 
 # rubocop:disable PrisonGuard/PrivateModule
 (foo(
-  # rubocop:enable PrisonGuard/PrivateModule
   foo
 ))
-  .flatten
+  .flatten # rubocop:enable PrisonGuard/PrivateModule
```
froydnj: This kind of change is not desirable; the comment should not be moved from its own line to an inline comment.

froydnj (changes requested):
(I don't feel qualified to weigh in on the major directional change this would be, just noting a problem with the proposed change.)

fixtures/small/paren_expr_calls_expected.rb
thomasklemm: True, standalone directives should stay standalone and not move inline

thomasklemm:
Thanks for the review @froydnj! You're right — fixed in af47f08.

Inside delimiter expressions (argument lists, arrays), the inline comment was drifting to the closing bracket because `flush_inline_comments` only fires at the next `HardNewLine`. Now these cases fall back to the existing behavior (comment moved above the code), and inline handling is limited to top-level statements and block bodies where it works correctly.

fables-tales:
I do not think we want to accept or allow this style.

PR #888, opened 2026-05-12
[fix] Render implicit nodes in hash rocket conversions

### LINKED ISSUE #884: Incorrectly formats hash with mixed value omission shorthand and hash rockets
**Version:** 0.13.9-6

[→ View on rubyfmt.run](https://rubyfmt.run/#eCA9IDEKeSA9IDprZXkKeyB4OiwgeSA9PiAyIH0=)

## Input

```ruby
x = 1
y = :key
{ x:, y => 2 }
```

## Output

```ruby
x = 1
y = :key
{:x =>, y => 2}
```

## Expected behavior

Correctly handle hashes that mix hash value omission shorthand (`x:,`) and hash rockets (`y => 2`).
For example:

```ruby
x = 1
y = :key
{x:, y => 2}
```

---

reese:
Closes #884 

For hashes that use the hash rocket syntax, the general style is to try and make all the keys match, so if any key uses hash rockets, all the keys will be converted to hash rockets for uniformity. In the case of implicit nodes, there's no AST node for the value, so when they're mixed with hash rockets (e.g. `{ foo:, bar => baz }`) we need to also re-render the value for that key after the rocket.

### DIFF
--- a/fixtures/small/hash_mixed_shorthand_and_rocket_actual.rb
+++ b/fixtures/small/hash_mixed_shorthand_and_rocket_actual.rb
@@ -1,0 +1,15 @@
+x = 1
+y = :key
+
+single_line = { x:, y => 2 }
+
+multi_line = {
+  x:,
+  y => 2
+}
+
+leading_rocket = { y => 2, x: }
+
+with_other_pairs = { a: 1, x:, y => 2 }
+
+with_splats = { x:, y => 2, **z}
--- a/fixtures/small/hash_mixed_shorthand_and_rocket_actual.rb
+++ b/fixtures/small/hash_mixed_shorthand_and_rocket_actual.rb
@@ -13,3 +13,5 @@ leading_rocket = { y => 2, x: }
 with_other_pairs = { a: 1, x:, y => 2 }
 
 with_splats = { x:, y => 2, **z}
+
+only_shorthands = {x:, y:}
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -2686,9 +2686,11 @@ fn format_assoc_node<'src>(ps: &mut ParserState<'src>, assoc_node: prism::AssocN
 
         let value = assoc_node.value();
         if let Some(implicit) = value.as_implicit_node() {
-            // A mixed hash forces hash rocket form, so we have to expand the
-            // shorthand `x:` into the equivalent `:x => x` rather than
-            // emitting nothing for the implicit value.
+            // If there's an implicit node here, that means we're in a hash shorthand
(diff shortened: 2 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2688,11 +2683,20 @@ fn format_assoc_node<'src>(ps: &mut ParserState<'src>, assoc_node: prism::AssocN
...
+
+        let value = assoc_node.value();
+        if let Some(implicit) = value.as_implicit_node() {
+            // A mixed hash forces hash rocket form, so we have to expand the
+            // shorthand `x:` into the equivalent `:x => x` rather than
+            // emitting nothing for the implicit value.
+            if !as_symbol {
+                ps.emit_space();
+                format_node(ps, implicit.value());
+            }
```
froydnj, line 2697: This comment feels like it doesn't line up with the code -- like the conditional should be a debug assertion or something?  Or that we didn't setup the `as_symbol` condition correctly or something

froydnj (approved):
I think this is good thing to fix, but I kind of wish that we were keeping the implicit value even in hash rocket form -- i.e. fix the implicit value to not emit the hash rocket in the first place.

Not going to block this, but putting it out there for consideration.

reese:
FWIW I'd be fine with keeping implicit nodes implicit (and making the output of `{ foo:, bar => baz }` unchanged), as it currently stands the PR is behaving as if it were `{ foo: foo, bar => baz }`, but I don't feel strongly that it has to be that way. (My personal opinion is mostly that I ~never use the hash rocket syntax and would be fine not doing this whole transformation of symbol keys into rockets in the first place, but I think we decided on the transformation a long time ago 🤷 )

cc @fables-tales if you've got any thoughts

librubyfmt/src/format_prism.rs
reese: I don't think it would be a debug assertion, because there's still the normative case where we run this branch in something like `{foo:, bar: baz}` where `as_symbol = true` and we do nothing. (But I can still definitely make the comment a bit clearer, I think `as_symbol` as a name might just be mildly confusing with this change.)

fixtures/small/hash_mixed_shorthand_and_rocket_actual.rb
```diff
@@ -0,0 +1,15 @@
+x = 1
+y = :key
+
+single_line = { x:, y => 2 }
```
fables-tales, line 4: can we make sure that in an all  implicit hash {x:, y:} it stays the same?

fables-tales (changes requested):

fables-tales (approved):

PR #891, opened 2026-05-18
[fix] Dedup parens for capitalized calls with only block args

### LINKED ISSUE #890: capitalized methods with receivers with passed blocks format as invalid Ruby
**Commit:** [`a0ebab0`](https://github.com/fables-tales/rubyfmt/commit/a0ebab0cfae44343b253814563ab6ffc39ab0d26)

#### Input

Reduced testcase from Stripe's codebase.

[→ View on rubyfmt.run](https://rubyfmt.run/#Y2xhc3MgRXhhbXBsZQogIGRlZiBzZWxmLmVzY2FwZSgmYmxrKQogICAgSGVscGVycy5FU0NBUEVfZm9yX3RyYW5zbWlzc2lvbigmYmxrKQogIGVuZAplbmQK)

```ruby
class Example
  def self.escape(&blk)
    Helpers.ESCAPE_for_transmission(&blk)
  end
end

```

#### Output

```ruby
class Example
  def self.escape(&blk)
    Helpers.ESCAPE_for_transmission()(&blk)
  end
end

```

This is invalid Ruby syntax -- pasting it back into `rubyfmt.run` will give parsing errors.

#### Expected behavior

The output should be valid Ruby.

I think this is a regression from #870 -- I did not think through all the possible combinations here, sorry. :(

reese:
Closes #890
(and fixing something introduced in #870)

On the tin -- block arguments in Prism are confusingly separate from the rest of the argument list, so we also need to handle cases where they're the only caller. Lone block args [have their own paren handling](https://github.com/fables-tales/rubyfmt/blob/3310e525448511cd02045bc2fd05feaa0ecbdbee/librubyfmt/src/format_prism.rs#L2036-L2045), so we should "elide" them in `use_parens_for_call_node` so that they're not rendered twice.

### DIFF
--- a/fixtures/small/constant_path_callers_actual.rb
+++ b/fixtures/small/constant_path_callers_actual.rb
@@ -24,3 +24,10 @@ Test.Case() { }
 Test.Case() do; end
 Test.Case(args) { }
 Test.Case(args) do ;end
+
+Case(&blk)
+Case(args, &blk)
+Test::Case(&blk)
+Test::Case(args, &blk)
+Test.Case(&blk)
+Test.Case(args, &blk)
--- a/fixtures/small/constant_path_callers_expected.rb
+++ b/fixtures/small/constant_path_callers_expected.rb
@@ -35,3 +35,10 @@ end
 Test.Case(args) { }
 Test.Case(args) do
 end
+
+Case(&blk)
+Case(args, &blk)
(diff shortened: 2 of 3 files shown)


froydnj (approved):

froydnj:
Thank you!

PR #894, opened 2026-05-22
[fix] Render inline conditionals in source order to fix line handling

### LINKED ISSUE #883: Surplus line break without indent on class level method call
**Commit:** [`1be9136`](https://github.com/fables-tales/rubyfmt/commit/1be913684df5d3224e3fc196d0bad5897d56b334)

#### Input

[→ View on rubyfmt.run](https://rubyfmt.run/#bW9kdWxlIEV4YW1wbGUKICBjbGFzcyBBcHBsaWNhdGlvbiA8IFJhaWxzOjpBcHBsaWNhdGlvbgogICAgY29uZmlnLm1pZGRsZXdhcmUuaW5zZXJ0X2JlZm9yZSgKICAgICAgQWN0aW9uRGlzcGF0Y2g6OkNvb2tpZXMsCiAgICAgIFNlc3Npb25Db29raWVVcGdyYWRlcgogICAgKSBpZiBSYWlscy5lbnYuc3RhZ2luZz8gfHwgUmFpbHMuZW52LnByb2R1Y3Rpb24/CiAgZW5kCmVuZA==)

```ruby
module Example
  class Application < Rails::Application
    config.middleware.insert_before(
      ActionDispatch::Cookies,
      SessionCookieUpgrader
    ) if Rails.env.staging? || Rails.env.production?
  end
end
```

#### Output

```ruby
module Example
  class Application < Rails::Application
    if 
Rails.env.staging? || Rails.env.production?
      config.middleware.insert_before(ActionDispatch::Cookies, SessionCookieUpgrader)
    end
  end
end

```

#### Expected behavior

When feeding the output back to rubyfmt for input, it formats it correctly and that is the expected output for the first run:
```ruby
module Example
  class Application < Rails::Application
    if Rails.env.staging? || Rails.env.production?
      config.middleware.insert_before(ActionDispatch::Cookies, SessionCookieUpgrader)
    end
  end
end
```
https://rubyfmt.run/#bW9kdWxlIEV4YW1wbGUKICBjbGFzcyBBcHBsaWNhdGlvbiA8IFJhaWxzOjpBcHBsaWNhdGlvbgogICAgaWYgClJhaWxzLmVudi5zdGFnaW5nPyB8fCBSYWlscy5lbnYucHJvZHVjdGlvbj8KICAgICAgY29uZmlnLm1pZGRsZXdhcmUuaW5zZXJ0X2JlZm9yZShBY3Rpb25EaXNwYXRjaDo6Q29va2llcywgU2Vzc2lvbkNvb2tpZVVwZ3JhZGVyKQogICAgZW5kCiAgZW5kCmVuZAo=

reese:
Closes #883 

Inline conditionals (like `foo if bar`) are a special type of breakable where we separately render the body and predicate so we can potentially render them as the multiline form if needed. When we do this, we need to make sure render those tokens in source order (body first, then predicate) so that the line numbers work correctly. This only really matters when we're rendering a multiline body, e.g.

```ruby
foo(
  a,
  b
) if bar
```

### DIFF
--- a/librubyfmt/src/comment_block.rs
+++ b/librubyfmt/src/comment_block.rs
@@ -54,37 +54,27 @@ impl CommentBlock {
         })
     }
 
+    /// Set each comment's leading indent to exactly `indent_depth` spaces
     pub fn apply_spaces(mut self, indent_depth: ColNumber) -> Self {
-        if indent_depth == 0 {
-            return self;
-        }
-
-        let indent = get_indent(indent_depth as usize);
+        let target = indent_depth as usize;
         for comment in &mut self.comments {
             // Ignore empty vecs -- these represent blank lines between
             // groups of comments
-            if !comment.is_empty() && !comment.starts_with(b"=begin") {
-                comment.to_mut().splice(0..0, indent.iter().copied());
-            }
-        }
-        self
-    }
-
-    /// Replace any baked-in leading space indent with `indent_depth` spaces.
-    /// Use this when a CommentBlock was captured under one indent and needs
-    /// to be emitted under another.
-    pub fn reindent_to(mut self, indent_depth: ColNumber) -> Self {
-        for comment in &mut self.comments {
             if comment.is_empty() || comment.starts_with(b"=begin") {
                 continue;
             }
-            let leading = comment.iter().take_while(|&&b| b == b' ').count();
-            if leading > 0 {
-                let trimmed = comment[leading..].to_vec();
-                *comment = Cow::Owned(trimmed);
+            let current = comment.iter().take_while(|&&b| b == b' ').count();
+            if current == target {
+                continue;
+            }
+            if current > target {
+                comment.to_mut().drain(0..(current - target));
+            } else {
+                let extra = get_indent(target - current);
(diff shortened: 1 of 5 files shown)


librubyfmt/src/comment_block.rs
```diff
@@ -70,6 +70,23 @@ impl CommentBlock {
...
+    /// to be emitted under another.
+    pub fn reindent_to(mut self, indent_depth: ColNumber) -> Self {
+        for comment in &mut self.comments {
+            if comment.is_empty() || comment.starts_with(b"=begin") {
+                continue;
+            }
+            let leading = comment.iter().take_while(|&&b| b == b' ').count();
+            if leading > 0 {
+                let trimmed = comment[leading..].to_vec();
+                *comment = Cow::Owned(trimmed);
```
froydnj: Is it possible that we could just manipulate the slice directly here by shifting the owned slice down, rather than allocating a separate vec?

It would be even cooler if the dedenting here could be automatically combined with the indenting, so we wouldn't necessary need changes if everything still lived at the same indent depth.

froydnj (approved):
Seems reasonable, non-blocking comments below.

PR #895, opened 2026-05-26
[fix] Only push start-of-file comments on empty lines

### LINKED ISSUE #889: Trailing comments on the first line of a file cause incorrect shifting
**Commit:** [`4f7d5f3`](https://github.com/fables-tales/rubyfmt/commit/4f7d5f33953eddc3bc56eeb1affd07faf3e0ffaa)

Follow up to #886, addressed in #887.

The first end-of-line comment is now correctly positioned, but subsequent end-of-line comments are incorrectly indented.

#### Input

[→ View on rubyfmt.run](https://rubyfmt.run/#Y2FzZSAwCmluIDEgIyBJbiAxCiAgZm9vCmluIDIgIyBJbiAyCiAgYmFyCmVuZCAjIEVuZA==)

```ruby
case 0
in 1 # In 1
  foo
in 2 # In 2
  bar
end # End
```

#### Output

```ruby
case 0
# In 1
in 1
  foo
  # In 2
in 2
  bar
  # End
end
```

Note that the same is true for `when` ([rubyfmt.run](https://rubyfmt.run/#Y2FzZSAwCndoZW4gMSAjIEluIDEKICBmb28Kd2hlbiAyICMgSW4gMgogIGJhcgplbmQgIyBFbmQ=)).

While writing the test case, I also noticed that empty branches cause strange behavior ([rubyfmt.run](https://rubyfmt.run/#Y2FzZSAxICMgQ2FzZQppbiAxICMgSW4gMQppbiAyICMgSW4gMgogICMgRm9vCiAgMgplbmQgIyBFbmQ=)):

#### Input

```ruby
case 1 # Case
in 1 # In 1
in 2 # In 2
  # Foo
  2
end # End
```

#### Output

```ruby
# Case
# In 1
# In 2
# Foo
case 1
in 1
in 2
  2
  # End
end
```

Even stranger ([rubyfmt.run](https://rubyfmt.run/#Y2FzZSAwICMgQ2FzZQppbiAxICMgSW4gMQppbiAyICMgSW4gMgplbmQgIyBFbmQKZm9vICMgRm9vCgpjYXNlIDAgIyBDYXNlCmluIDEgIyBJbiAxCmluIDIgIyBJbiAyCmVuZCAjIEVuZApmb28gIyBGb28=)):

#### Input

```ruby
case 0 # Case
in 1 # In 1
in 2 # In 2
end # End
foo # Foo

case 0 # Case
in 1 # In 1
in 2 # In 2
end # End
foo # Foo
```

#### Output

```ruby
# Case
# In 1
# In 2
# End
# Foo
case 0
in 1
in 2
end

foo

# Case
case 0
# In 1
in 1
  # In 2
in 2
  # End
end
# Foo
foo
```

reese:
Closes #889 

The example in that issue is a bit of a red herring. Specifically, the issue only reproduces when a trailing comment is also the first line of a file, where we treat them as start-of-line comments (like `# frozen_string_literal: true`, `# typed: strict` etc.) and try and collect all of them up front. We should only do that when the comment is the only thing on the line and should skip doing that when it's trailing other code.

### DIFF
--- a/fixtures/small/prism/case_match_empty_branches_with_comments_actual.rb
+++ b/fixtures/small/prism/case_match_empty_branches_with_comments_actual.rb
@@ -1,0 +1,18 @@
+case 1 # Case
+in 1 # In 1
+in 2 # In 2
+  # Foo
+  2
+end # End
+
+case 0 # Case
+in 1 # In 1
+in 2 # In 2
+end # End
+foo # Foo
+
+case 0 # Case
+when 1 # When 1
+when 2 # When 2
+end # End
+bar # Bar
--- a/fixtures/small/prism/case_match_empty_branches_with_comments_expected.rb
+++ b/fixtures/small/prism/case_match_empty_branches_with_comments_expected.rb
@@ -1,0 +1,32 @@
+# Case
+case 1
+# In 1
+in 1
+  # In 2
+in 2
+  # Foo
+  2
+  # End
+end
+
+# Case
+case 0
+# In 1
+in 1
+  # In 2
+in 2
+  # End
+end
+# Foo
+foo
+
+# Case
+case 0
+# When 1
+when 1
+  # When 2
+when 2
+  # End
+end
+# Bar
+bar
--- a/librubyfmt/src/file_comments.rs
+++ b/librubyfmt/src/file_comments.rs
@@ -95,6 +95,8 @@ impl FileComments {
         let line_index = LineIndex::from_vec(line_starts);
 
         let mut file_comments = FileComments::default();
+        file_comments.lines_with_ruby = lines_with_ruby;
+
         for comment in comments {
             file_comments.push_comment(
                 line_index.get_line_number(comment.location().start_offset()) as u64,
@@ -105,7 +107,6 @@ impl FileComments {
                 .push(comment.location().start_offset());
         }
 
-        file_comments.lines_with_ruby = lines_with_ruby;
         file_comments.last_lineno = line_index.line_starts.len() as u64;
         file_comments.line_index = line_index;
         file_comments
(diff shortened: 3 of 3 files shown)


nicholasdower:
fwiw, it also happens when there isn't anything separating the comments at the beginning of the file. Like ([rubyfmt.run](https://rubyfmt.run/#IyBUaGlzIGlzIGEgc2NyaXB0LgpjYXNlIEFSR1ZbMF0gIyBDaGVjayB0aGUgZmlyc3QgYXJnCmluICcxJyAjIEl0IHdhcyBhIDEKICBwdXRzIDEKZWxzZQogIHB1dHMgIm90aGVyIgplbmQ=)):
```ruby
# This is a script.
case ARGV[0] # Check the first arg
in '1' # It was a 1
  puts 1
else
  puts "other"
end
```

```ruby
# This is a script.
# Check the first arg
# It was a 1
case ARGV[0]
in "1"
  puts(1)
else
  puts("other")
end
```

reese:
```
# This is a script.
```
in that example is still a comment at the start of the file, that's what causes the bug to happen. Even just [having a newline](https://rubyfmt.run/#CiMgVGhpcyBpcyBhIHNjcmlwdC4KY2FzZSBBUkdWWzBdICMgQ2hlY2sgdGhlIGZpcnN0IGFyZwppbiAnMScgIyBJdCB3YXMgYSAxCiAgcHV0cyAxCmVsc2UKICBwdXRzICJvdGhlciIKZW5k) up top means it doesn't happen. All that matters is that consecutive lines at the start of the file have comments, so it also doesn't repro if [there's not comments on contiguous lines](https://rubyfmt.run/#IyBUaGlzIGlzIGEgc2NyaXB0LgpjYXNlIEFSR1ZbMF0KaW4gJzEnICMgSXQgd2FzIGEgMQogIHB1dHMgMQplbHNlCiAgcHV0cyAib3RoZXIiCmVuZA==).

The fix in this PR works for the example you gave (unless I'm misreading your comment and that's still somehow different, but I ran it locally and looks as-expected).

froydnj (approved):

PR #896, opened 2026-05-26
[fix] Multiline call chains with multi-statement blocks

### LINKED ISSUE #879: Multi-expression brace blocks in call chains are not idempotent
**Commit:** [`94566e5`](https://github.com/fables-tales/rubyfmt/commit/94566e5dfc23d4dbcf73bf643313612392920038)

#### Input

[→ View on rubyfmt.run](https://rubyfmt.run/#c29tZV9hcnJheS5tYXAgeyB8eHwgYTsgYiB9LnNlbGVjdCB7ICB9Cgpzb21lX2FycmF5Lm1hcCB7IHx4fAogIGEKICBiCn0uc2VsZWN0IHsgfQ==)

```ruby
# First pass
some_array.map { |x| a; b }.select {  }

# Second pass
some_array.map { |x|
  a
  b
}.select { }
```

#### Output

```ruby
# First pass
some_array.map { |x|
  a
  b
}.select { }

# Second pass
some_array
  .map { |x|
    a
    b
  }
  .select { }

```

#### Expected behavior

We should get to the second output on the first try

reese:
Closes #879 

This fixes an idempotency issue with call chains containing multi-statement blocks (like `foo { a; b }`). We need to explicitly check for these in the multilining logic since they aren't breakables and always multiline.

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -2270,9 +2270,9 @@ fn format_call_chain_segments<'src>(
             })
         });
 
-        let is_user_multilined = call_chain_elements_are_user_multilined(ps, &chain_elements);
+        let should_multiline = call_chain_should_be_multilined(ps, &chain_elements);
 
-        ps.breakable_call_chain_of(is_user_multilined, |ps| {
+        ps.breakable_call_chain_of(should_multiline, |ps| {
             // Recurse and format previous segments inside this breakable
             format_call_chain_segments(ps, segments);
 
@@ -2426,10 +2426,7 @@ fn format_call_body<'src>(
     }
 }
 
-fn call_chain_elements_are_user_multilined(
-    ps: &ParserState,
-    call_chain_elements: &[prism::Node],
-) -> bool {
+fn call_chain_should_be_multilined(ps: &ParserState, call_chain_elements: &[prism::Node]) -> bool {
     // Making a mutable copy since we may pop some items off later
     let mut call_chain_elements = call_chain_elements;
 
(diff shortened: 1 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -2514,18 +2513,50 @@ fn call_chain_elements_are_user_multilined(
...
+
+    false
+}
+
+fn block_body_renders_multiline(opening: &[u8], body: Option<prism::Node>) -> bool {
+    if opening == b"do" {
+        return true;
+    }
+    body.and_then(|n| n.as_statements_node())
+        .is_some_and(|statements| statements.body().len() > 1)
```
froydnj, line 2557: Is this going to get us in trouble if a block has a single very long send where the send gets multilined, but we think the block is fine?

froydnj (approved):
Just one question about the test coverage here.

librubyfmt/src/format_prism.rs
reese: What's an example? Do you mean something like

```ruby
foo { bar("assume this is string is like 120 chars or something...") }
```

I think if the send inside the block is sufficiently long for that to happen, the block itself would also be over the maximum line length.

froydnj: Yeah, that's approximately what I was thinking, or `Some::Really::Long::Constant.sendsit`.

idk, maybe there are weird cases where the block and/or send are not long enough, but are required to be in a position where they would overflow?  We can just kick the can down the road, though.

PR #907, opened 2026-06-02
Simplifies installation into vim with a plugin

robacarp:
Hi there! 

I've tweaked the vim integration a small amount just to allow it to be installed from the git repo instead of needing to clone locally. I've updated the readme instructions as well. Now, pointing this at my fork, I'm able to install the rubyfmt vim extension without manually sourcing the config and overriding the path. 

I'm able to:

- `brew install rubyfmt`
- and add this to my vim-plugin list: `Plug 'robacarp/rubyfmt', { 'rtp': 'editor_plugins/vim', 'branch': 'vim-plugin' }`

And I have a working configuration. Hopefully this contribution is welcome, I believe it'll make using the tool much easier for vim-natives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,20 @@ See the [null-ls documentation](https://github.com/jose-elias-alvarez/null-ls.nv
 
 ### Vim
 
-1. Run `cargo build --release`
-2. Add `source /path/to/rubyfmt.vim` to your `~/.vimrc`
-3. Add `let g:rubyfmt_path = /path/to/target/release/rubyfmt-main` beneath the source line
+vim-plug:
+
+```vim
+Plug 'fables-tales/rubyfmt', { 'rtp': 'editor_plugins/vim' }
+```
+
+Native packages:
+
+```sh
+git clone https://github.com/fables-tales/rubyfmt ~/.rubyfmt
+ln -s ~/.rubyfmt/editor_plugins/vim ~/.vim/pack/rubyfmt/start/rubyfmt
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -127,9 +127,20 @@ See the [null-ls documentation](https://github.com/jose-elias-alvarez/null-ls.nv
...
-1. Run `cargo build --release`
-2. Add `source /path/to/rubyfmt.vim` to your `~/.vimrc`
-3. Add `let g:rubyfmt_path = /path/to/target/release/rubyfmt-main` beneath the source line
+vim-plug:
+
+```vim
+Plug 'fables-tales/rubyfmt', { 'rtp': 'editor_plugins/vim' }
+```
+
+Native packages:
```
fables-tales, line 136: I don't know about vim native packages, does this work for users who do not use plug?

robacarp: I'm by no means a vim-plugin expert. My understanding is that tools like vim-plug, pathogen, etc, all handle the "fetching of code and putting it in place" while vim itself handles loading of the plugin through the autoload system or the packs system. Using vim-plug or pathogen benefits the user in that they're able to run one command and take updates to all the packages installed.

So the instructions for installing without a plugin tool are simply to download the plugin itself and place it in the  correct place. The formatting and placement of the files within the plugin is standardized by vim itself. 

That said, I tested these instructions both with the plug pattern and the native package and saw the tool work correctly in both cases.

fables-tales (approved):

PR #910, opened 2026-06-07
Replace direct `regex` crate dependency

reese:
We currently just have a single usage of the `regex` crate, which is checking for `# rubyfmt: (true|false)` magic comments. Using regex for this is a bit overkill, and we can instead use a plain ol' string scan, which (1) is a good bit faster, at least from a quick local benchmark and (2) doesn't require any external dependencies.


froydnj (approved):
Fewer dependencies is more better.

Issue #911, opened 2026-06-07
Syntax Error: Inline Comments after `rescue \`


nicholasdower:
**Commit:** [`95210d7`](https://github.com/fables-tales/rubyfmt/commit/95210d780317a4920ac6aae9c0a7d5730c0739d1)

#### Input

[→ View on rubyfmt.run](https://rubyfmt.run/#YmVnaW4KcmVzY3VlIFwKICAgIEZvbywgIyBjb21tZW50CiAgICBCYXIKZW5kCg==)

```ruby
begin
rescue \
    Foo, # comment
    Bar => ex
  puts ex
end
```

#### Output

```ruby
begin
rescue  # comment
 Foo,
  Bar => ex
  puts(ex)
end
```

#### Expected behavior

The comment should be moved above `rescue` and `Foo` should be moved to the `rescue` line.

reese:
Closed by #912

PR #912, opened 2026-06-07
[fix] Handle inline comments after `rescue \`

nicholasdower:
Attempt at closing #911

### DIFF
--- a/librubyfmt/src/format_prism.rs
+++ b/librubyfmt/src/format_prism.rs
@@ -4728,6 +4728,7 @@ fn format_rescue_node<'src>(ps: &mut ParserState<'src>, rescue_node: prism::Resc
     let exceptions = rescue_node.exceptions();
     let reference = rescue_node.reference();
     if !exceptions.is_empty() {
+        // Preemptively extract inline comments so they land before rescue, not inside the breakable.
         ps.at_offset(exceptions.first().unwrap().location().start_offset());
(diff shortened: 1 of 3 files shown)


librubyfmt/src/format_prism.rs
```diff
@@ -4728,6 +4728,7 @@ fn format_rescue_node<'src>(ps: &mut ParserState<'src>, rescue_node: prism::Resc
     let exceptions = rescue_node.exceptions();
     let reference = rescue_node.reference();
     if !exceptions.is_empty() {
+        ps.at_offset(exceptions.iter().next().unwrap().location().start_offset());
```
reese: This makes sense, but you may want to provide some rationale for why this is here since it's not particularly obvious at first glance. E.g. `// Preemptively extract comments from the first exception that may need to render outside the breakable` or something to that effect.

reese:
```suggestion
        ps.at_offset(exceptions.first().unwrap().location().start_offset());
```

or perhaps even better yet

```rust
if let Some(exception) = exceptions.first() {
  ps.at_offset(exception.location().start_offset());
```

to avoid the `unwrap`

nicholasdower: Makes sense. Thank you.

reese (approved):

PR #951, opened 2026-07-27
Enforce at least one space after comment symbol

aadi-stripe:
Autoformats comments to have at least one space after `#` for better readability, i.e.:

```
#comment
```

will be formatted to:
```
# comment
```

Note: the space is added after the first non-`#` leading character, i.e.:
```
###comment
```

✅ will be formatted to:
```
### comment
```

❌ **and not to:**
```
# ##comment
```

A few exceptions will be untouched, since these hold special meaning or count as convention:
```
#!
#-
#=
#:
```


fables-tales:
@reese peaceful?

reese:
Peaceful with the concept, there's still some stuff to fix in the PR (which is why I assume it's a draft), such as exempting shebang comments from this (I don't think something like `# !/usr/bin/env ruby` works or is preferable) but I'm cool with this as a thing for regular comments

aadi-stripe:
> Peaceful with the concept, there's still some stuff to fix in the PR (which is why I assume it's a draft), such as exempting shebang comments from this (I don't think something like `# !/usr/bin/env ruby` works or is preferable) but I'm cool with this as a thing for regular comments

Yep, still working on this

jez:
Another load bearing sigil is `#:` for type annotations

https://sorbet.org/docs/rbs-support

librubyfmt/src/comment_block.rs
```diff
@@ -54,6 +61,44 @@ impl CommentBlock {
...
+                    .take_while(|&&b| b == b'#')
+                    .count();
+
+                let anchor = start_idx + collapse;
+                if anchor == comment.len() {
+                    // No point truncating an empty comment.
+                    continue;
+                }
+
+                let current = comment[anchor..].iter().take_while(|&&b| b == b' ').count();
```
reese: nit: if we only care about this count being > 0, it might be nice to end this iter early or not iter at all. It doesn't seem like we use `current` anywhere after this, so it might just be as simple as

```rust
if comment[anchor].is_ascii_whitespace() { continue; } 
```

or something to that effect.

librubyfmt/src/comment_block.rs
```diff
@@ -54,6 +61,44 @@ impl CommentBlock {
...
+                    continue;
+                }
+
+                let next_char = comment[anchor];
+                if SPECIAL_CHARS_AFTER_HASH_TO_IGNORE.contains(&next_char) {
+                    continue;
+                }
+
+                // Add at least one space
+                comment.to_mut().splice(anchor..anchor, [b' ']);
```
reese: We're only ever inserting exactly one space at the front here, right? I think `insert` might be more efficient in that case.


```suggestion
                // Add one space
                comment.to_mut().insert(anchor, b' ');
```

reese:
LGTM besides the nits above. I'm gonna leave this for Stripe folks (r? @fables-tales || @froydnj) to give final approval to make sure it aligns with whatever style expectations y'all have agreed on.

librubyfmt/src/comment_block.rs
aadi-stripe: Ah sure

librubyfmt/src/comment_block.rs
aadi-stripe: good call

aadi-stripe:
> LGTM besides the nits above. I'm gonna leave this for Stripe folks (@fables-tales || @froydnj) to give final approval to make sure it aligns with whatever style expectations y'all have agreed on.

@reese Thank you! And also, 👋

librubyfmt/src/comment_block.rs
```diff
@@ -12,6 +12,13 @@ pub struct CommentBlock {
     comments: Vec<Cow<'static, [u8]>>,
 }
 
+const SPECIAL_CHARS_AFTER_HASH_TO_IGNORE: [u8; 4] = [
+    b'!', // #! 'shebang' comments cannot be modified since they are Unix directives.
+    b'=', // #=== is a common delimiting pattern.
+    b'-', // #--- is a common delimiting pattern.
+    b':', // #: is used for RBS directives.
```
reese: Oh one last thing: it might be worth checking all the options/exceptions for the [Layout/LeadingCommentSpace](https://www.rubydoc.info/gems/rubocop/0.85.1/RuboCop/Cop/Layout/LeadingCommentSpace) cop, it has some other things we don't handle here like `#**` for Doxygen comments, `#ruby=2.7.0` directives in Gemfiles, etc. that we don't handle here.

Definitely should at least add `#|`, which is also used for RBS support for sigs.

aadi-stripe: ah sure

froydnj (approved):

